PR #2, opened 2017-06-05
Added instructions to start smtp debugging server

rohithasrk:


README.rst
```diff
@@ -100,6 +106,18 @@ Setup (integrate in an existing django project)
 
     urlpatterns += staticfiles_urlpatterns()
 
+Testing (locally)
```
nemesisdesign: Testing can be confused with how to run automated tests, could you replicate the ["Installing for development" section of django-netjsonconfig](https://github.com/openwisp/django-netjsonconfig/#installing-for-development)?

README.rst
```diff
@@ -85,6 +85,12 @@ Setup (integrate in an existing django project)
         'openwisp_users',
     ]
 
+Set ``EMAIL_PORT`` in ``settings.py`` to a port number. For example:
```
nemesisdesign: If put here it could be confused as a mandatory setting, while this would be normally done only during development, therefore could you move this suggestion to "Testing (locally)" (which I also suggest to rename, see other comment).

nemesisdesign:
thx @rohithasrk 👍

PR #14, opened 2017-12-18
[ci] Added django 2.0 to tests

AZtheAsian:

PR #15, opened 2017-12-19
adding django2.0 compatibility to openwisp-users

gandhi23:
Adding django2.0 compatibility to openwisp-users and thus, removing older django versions compatibility

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,13 +2,14 @@ language: python
 sudo: false
 cache: pip
 
-python:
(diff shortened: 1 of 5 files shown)


nemesisdesign:
Sloved by #14, I don't know why you started working on this task even though it was already claimed by another student. I suggest you to not do that otherwise you risk to end up in the actual situation. It's still good for you because you learn but you won't score tasks.

PR #25, opened 2018-07-27
update readme integration section

AlmogCohen:
Add some missings steps and information into the README integration section to make it work more fluently for new users


AlmogCohen:
Funny story - I was trying to use it within another project just because it
tied up the organizations and allauth together. After playing with it a
little I've decided going with something else as my needs have changed. You
deserved the PR any way for the good work :)

PR #26, opened 2018-07-30
[admin] Fixed Integrity error if trying to change email that exists #23

### LINKED ISSUE #23: [admin] IntegrityError if trying to change an email that already exists
- create an user
- create a second user with a different email address
- change the email of the second user to the email of the first user

Expected result: validation error in UI
Actual result: 500 internal server error (`IntegrityError`)

R9295:
Fixes and closes #23

### DIFF
--- a/openwisp_users/admin.py
+++ b/openwisp_users/admin.py
@@ -6,6 +6,7 @@ from django.contrib.auth.admin import GroupAdmin as BaseGroupAdmin
 from django.contrib.auth.admin import UserAdmin as BaseUserAdmin
 from django.contrib.auth.forms import UserChangeForm as BaseUserChangeForm
 from django.contrib.auth.forms import UserCreationForm as BaseUserCreationForm
+from django.core.exceptions import ValidationError
 from django.utils.translation import ugettext_lazy as _
 from organizations.base_admin import (BaseOrganizationAdmin,
                                       BaseOrganizationOwnerAdmin,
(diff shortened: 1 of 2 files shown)


R9295:
I think I know why the coverage is decreasing, will re-implement.

R9295:
@nemesisdesign should be fixed.

openwisp_users/admin.py
```diff
@@ -41,6 +42,24 @@ class OrganizationUserInline(admin.StackedInline):
 class EmailRequiredMixin(forms.ModelForm):
     email = forms.EmailField(label=_('Email'), max_length=254, required=True)
 
+    def _clean_email(self, email):
+        try:
+            EmailAddress.objects.get(email=email)
```
nemesisdesign: I'm not entirely sure here, think about this scenario:

- the user has a secondary address already existing in the system
- he modifies his email in the user form to match the secondary address

The email already exists so the validation will fail.

Another note: counting instead of retrieving the whole model instance will be faster, let's use that.

Could you instead try performing the following 2 checks?

1. no other user has this email address, eg: `User.objects.filter(email=email).count() < 1`
2. no other email address record for other users exists, eg: `EmailAddress.objects.filter(email=email).exclude(user=user).count() < 0` (you should be able to find the `user_id` somewhere in `self` or `self.cleaned_data`

openwisp_users/admin.py
```diff
@@ -41,6 +42,24 @@ class OrganizationUserInline(admin.StackedInline):
...
+
+    def clean(self):
+        cleaned_data = super(EmailRequiredMixin, self).clean()
+        # check if editing
+        try:
+            EmailAddress.objects.get(email=cleaned_data.get('email'))
+            if 'email' in self.changed_data:
+                self._clean_email(email=cleaned_data.get('email'))
+        except EmailAddress.DoesNotExist:
+            self._clean_email(email=cleaned_data.get('email'))
```
nemesisdesign: why are you performing identical operations to the ones performed in `_clean_email`?
Can't you just do:

```python
if 'email' in self.changed_data:
    self._clean_email(email=cleaned_data.get('email'))
```

openwisp_users/admin.py
R9295: Can't believe I missed that!

PR #30, opened 2018-09-16
[migrations] Set default organization

### LINKED ISSUE #29: [organizations] Create default organization with a data migration
We should create the default organization with a data migration.

Right now we rely on ansible-openwisp2 to do this but this solution implies the ansible role must be used. But openwisp2 or one of its derivative may be installed with a different strategy, having a data migration is a more robust solution.

There are a couple of catches though:

1. the default organization must be created only if no other organization already exists
2. we need to automatically (and randomly) generate the `uuid` for security reasons (this field can be used internally for things like authentication, eg: https://github.com/openwisp/openwisp-radius/issues/12

atb00ker:
Sets the default organisation with a data migration

- [x] the default organisation must be created only if no other organisation already exists
- [x] Automatically (and randomly) generate the UUID

Closes #29


openwisp_users/migrations/0003_default_organization.py
```diff
@@ -0,0 +1,42 @@
...
+
+def set_default_organization_uuid(apps, schema_editor):
+    """
+    Get default organzation or create a default
+    organization and set settings.DEFAULT_ORG_UUID
+    """
+    organization = apps.get_model('openwisp_users', 'organization')
+    try:
+        default_organization = organization.objects.get(
+            slug=settings.DEFAULT_ORG_SLUG)
```
atb00ker: @nemesisdesign 
For the case of user having multiple organizations,
it might be good to provide an option to enter the organization that user feels is appropriate to be default, hence by using `DEFAULT_ORG_SLUG`, the user can define the default organization.

Is this possibly useful or should i set default to the first organisation i get from database? Please suggest! :smile:

openwisp_users/migrations/0003_default_organization.py
```diff
@@ -0,0 +1,42 @@
...
+        )
+
+    settings.DEFAULT_ORG_SLUG = default_organization.slug
+    settings.DEFAULT_ORG_UUID = default_organization.pk
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('openwisp_users', '0002_auto_20180508_2017'),
```
atb00ker: The migration `('openwisp_users', '0002_auto_20180508_2017')` is a dependency in openwisp-ipam only as of now, since ipam has not been released is it possible to rename this dependency's name to something more descriptive?

nemesisdesign: unfortunately not, that was already released :(

openwisp_users/migrations/0003_default_organization.py
```diff
@@ -0,0 +1,42 @@
...
+        default_organization = organization.objects.create(
+            name='default',
+            slug='default',
+            is_active=False,
+            description='This is the default organization. '
+            'It created because no organization was set as default '
+            'during setup. If you want to delete it, please make sure '
+            'all the users are moved to appropriate organization '
+            'because when an organization is delete all the '
+            'organization users are deleted as well!',
```
nemesisdesign: I like this idea. I would change the hint and tell them that instead of deleting it they should simply rename it.

openwisp_users/migrations/0003_default_organization.py
nemesisdesign: unfortunately this was already released and I only noticed afterwards.
We could add a GCI task to write a script which checks that the last migration does not create `_auto_` and once we get it working in one of the django modules we can add it to all of them, what do you think?

openwisp_users/migrations/0003_default_organization.py
```diff
@@ -0,0 +1,40 @@
...
+from django.conf import settings
+
+
+def set_default_organization_uuid(apps, schema_editor):
+    """
+    Get default organzation or create a default
+    organization and set settings.DEFAULT_ORG_UUID
+    """
+    organization = apps.get_model('openwisp_users', 'organization')
+    try:
```
nemesisdesign: I think the first thing you have to do is to check whether any organization already exists, if it does just return and do nothing. This will avoid creating new default organizations in already existing openwisp instances.

openwisp_users/migrations/0003_default_organization.py
```diff
@@ -0,0 +1,40 @@
...
+    except (AttributeError, organization.DoesNotExist):
+        default_organization = organization.objects.create(
+            name='default',
+            slug='default',
+            is_active=False,
+            description='This is the default organization. '
+            'It was created because no organization was set as '
+            'default during setup. You can simply rename it '
+            'to your preferred organization name.'
+        )
```
nemesisdesign: Good idea. I'd replace `It was created because no organization was set as default during setup` with something like "it was created automatically at install time" or anything better.

openwisp_users/migrations/0003_default_organization.py
```diff
@@ -0,0 +1,40 @@
...
+    organization = apps.get_model('openwisp_users', 'organization')
+    try:
+        default_organization = organization.objects.get(
+            slug=settings.DEFAULT_ORG_SLUG)
+
+    except (AttributeError, organization.DoesNotExist):
+        default_organization = organization.objects.create(
+            name='default',
+            slug='default',
+            is_active=False,
```
nemesisdesign: why this is `False`??

openwisp_users/migrations/0003_default_organization.py
```diff
@@ -0,0 +1,40 @@
...
+    Get default organzation or create a default
+    organization and set settings.DEFAULT_ORG_UUID
+    """
+    organization = apps.get_model('openwisp_users', 'organization')
+    try:
+        default_organization = organization.objects.get(
+            slug=settings.DEFAULT_ORG_SLUG)
+
+    except (AttributeError, organization.DoesNotExist):
+        default_organization = organization.objects.create(
```
nemesisdesign: Please don't use `objects.create` as it skips the validation and `save` method of models, instantiate the object, call `full_clean` and `save`

openwisp_users/migrations/0003_default_organization.py
nemesisdesign: as in the previous comment, the organization should be created only if it doesn't already exist.
I'd rename the temporary setting to `_OPENWISP_DEFAULT_ORG_SLUG`, the leading underscore indicating it's a python convention to indicate an internal thing and shouldn't be relied upon.

openwisp_users/migrations/0003_default_organization.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('openwisp_users', '0002_auto_20180508_2017'),
+    ]
+
+    operations = [
+        migrations.RunPython(set_default_organization_uuid),
```
nemesisdesign: you need to add `, reverse_code=migrations.RunPython.noop` or it won't be possible to run the migrations backward

PR #35, opened 2018-10-01
[qa] Include migration file name test

atb00ker:
Include tests for checking if the migration file has a descriptive name, if the file has default name, fail ci.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,6 +25,9 @@ branches:
 before_install:
   - pip install --no-cache-dir -U -r requirements-test.txt
   - ./runflake8
+  - # TODO: remove after openwisp-utils > 0.2 is released
+  - pip install https://github.com/openwisp/openwisp-utils/tarball/master#egg=openwisp_utils --upgrade
+  - ./runcheckmigration
   - if [[ $TRAVIS_PYTHON_VERSION == 3* ]]; then ./runisort; fi
 
 # command to install requirements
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -26,6 +26,7 @@ before_install:
   - pip install --no-cache-dir -U -r requirements-test.txt
   - ./runflake8
   - if [[ $TRAVIS_PYTHON_VERSION == 3* ]]; then ./runisort; fi
+  - if [[ $TRAVIS_PYTHON_VERSION == 3* ]]; then ./runcheckmigration; fi
```
nemesisdesign: can you try removing the if and let the check happen on both python 2 and python 3? The previous line is a hack to avoid wasting too much time on differences between isort on python2 and 3 which have happened from time to time, but with the script you wrote this shouldn't happen right?

atb00ker:
@nemesisdesign 
For the other repositories, we can add the issue for GCI.
Since there are eight repositories, i think we can divide them in 2 task, please review the same. :smile: 

```
We need to write a new test to check if the migrations created have have been renamed to something more descriptive.

The following repositories need this test:

- openwisp-utils
- django-freeradius
- openwisp-ipam
- django-netjsonconfig

Example for implementation: https://github.com/openwisp/openwisp-users/pull/35/files
```
```
We need to write a new test to check if the migrations created have have been renamed to something more descriptive.

The following repositories need this test:

- django-x509
- django-ipam
- openwisp-radius
- django-loci


Example for implementation: https://github.com/openwisp/openwisp-users/pull/35/files
```

Issue #32, opened 2018-09-28
[admin] Allow non-superadmins to manage their own users


nemesisdesign:
Users which are not a superadmins but have permissions to edit users should work be able to do so in the following way:

- [x] should be able to see and edit users of the organizations they are member of
- [ ] should not be able to see `is_superuser` column in list
- [ ] should not be able to see `is_superuser` field in change form (now is readonly)
- [ ] should be able to add new users but only set them as members of their organization
- [ ] when editing, they shouldn't be able to set users as members of other organizations they are not member of, actually the organization select should only allow them to select one of their orgs

I started some of this in branch https://github.com/openwisp/openwisp-users/tree/issues/32 but I don't have time to finish all these items, help is greatly appreciated.

**EDIT**: follow up with list of things done and to do on https://github.com/openwisp/openwisp-users/pull/38

atb00ker:
1. Should an operator be allowed to create users without any organisation?
2. Should other organisations be visible to operators?
3. Should operator be allowed to see other organisation owners?
4. However users are not visible to operator from `User` list but still visible from `Organization users`, i think that's unintentional, right? i should implement `get_queryset` logic for `Organization user` as well, right?

nemesisdesign:
@atb00ker:

Premise, I forgot to tell you I found out this issue: https://github.com/openwisp/openwisp-utils/issues/14
I think it makes more sense to keep the multi-tenancy features (and their tests) in this module rather than openwisp-utils, keep in mind this when working on this issue because you may likely have to edit some of that code.

I was writing a very long reply but I wanted to shorten it: I think I'm either misunderstanding you or something may not be clear.
In this issue I want to focus on `openwisp_users.admin.UserAdmin`, the other admin classes like  `OrganizationAdmin`, `OrganizationUserAdmin`, `OrganizationOwnerAdmin` are restricted (at the moment I always do this manually but with #31 we'll start automating this), I mean usually operators cannot manipulate the models of those admin classes. We can and we should make also those admin classes multitenant but we can work on it in a separate issue or make it a GCI task because it's not very urgent.

f I misunderstood something or something is not clear let me know.

Issue #34, opened 2018-09-28
[admin] Only organizationuser objects which have admin=True should be able to manage their organization in the admin


nemesisdesign:
It could happen that a user which has `is_staff` set to `True` becomes member of another organization because he wants to use some service (implemented in other openwisp modules) of that organization as a user, but at the moment we can't let that happen because in the moment he become member of that organization he will also be able to edit the objects of that organization from the admin.

To solve this problem, we should change the code so that only `OrganizationUser` objects which are both member of an organization and have `is_admin` set to `True` are able to see and edit the objects of that organization.

When `OrganizationUser.is_admin` is set to `False` the user won't be allowed to see anything related to that organization in the admin.

OltarzewskiK:
Can I take this issue?

nemesisdesign:
@OltarzewskiK yes you can :-)

OltarzewskiK:
@nemesisdesign Ok, I will try to get down work in weekend, but gci is coming soon, so after beginning of google code in I would like to focus on contest's tasks, so Can I send PR (if I won't finish it before gci) after competition?

nemesisdesign:
@OltarzewskiK up to you, there are no requirements in open source, contributors contribute when they can. If someone else comes up and can finish this issue before you do we shouldn't prevent them from doing so, because it's more important to fix the issue than who fixes it, I'll send you some suggestions via PM

OltarzewskiK:
@nemesisdesign For doing this task should I use existing function for example [this](https://github.com/openwisp/openwisp-users/blob/master/openwisp_users/multitenancy.py#L27) or create new?

OltarzewskiK:
@nemesisdesign Are you sure that I have to change this method: https://github.com/openwisp/openwisp-users/blob/master/openwisp_users/admin.py#L195-L213 [the class](https://github.com/openwisp/openwisp-users/blob/master/openwisp_users/admin.py#L123) where it is referring to is I think `User`not `OrganizationUser`, so have I to create new class in the `admin`?

PR #38, opened 2018-10-08
[multi-tenancy] Hide content from operators #32

atb00ker:
**Note:** Changes need manual testing! Bugs are skipping through tests.

Hide other organisation users and and superusers from operators

**To be Moved to separate issue**
- **Bug:** if operator not admin in an organisation, the `USER ORGANIZATIONS` contains `-----` instead of the name of organisation. 

**Completed:**
- [x] should be able to see and edit users of the organizations they are member of
- [x] should not be able to see `is_superuser` column in list
- [x] should not be able to see `is_superuser` field in change form (now is readonly)
- [x] should be able to add new users but only set them as members of their organization
- [x] when editing, they shouldn't be able to set users as members of other organizations they are not member of, actually the organization select should only allow them to select one of their orgs
- [x] **Bug:** if not `is_staff` is given multiple organisations, an error is encountered.
- [x] Test cases 

Closes https://github.com/openwisp/openwisp-users/issues/32


openwisp_users/admin.py
```diff
@@ -139,10 +179,32 @@ def has_change_permission(self, request, obj=None):
...
-        # hide superusers from operators (they can't edit their details)
+        """
+        if operator is logged in - show only users
+        from same organization and hide superusers
+        if superuser is logged in - show all users
+        """
         if not request.user.is_superuser:
-            qs = qs.filter(is_superuser=False)
+            user = request.user
+            org_users = OrganizationUser.objects.filter(user=user, is_admin=True) \
```
atb00ker: if an operator `is_admin` in org1 and **not** `is_admin` in org2
then he should be able to see users of org1 only, isn't it?

The `get_queryset()` in previous commit only checks if `is_superuser`, then takes all his organisations and shows records of all the users in all the organisations.
I have added a check for `is_admin` as well. Is my understanding of `is_admin` correct?

openwisp_users/tests/test_admin.py
```diff
@@ -124,11 +115,12 @@ def test_admin_change_user_is_superuser_editable(self):
...
         self.assertContains(response, html)
 
-    def test_admin_change_user_is_superuser_readonly(self):
+    def test_admin_change_user_is_superuser_absent(self):
         operator = self._create_operator()
+        self._create_org_user()
         self.client.force_login(operator)
         response = self.client.get(reverse('admin:openwisp_users_user_change', args=[operator.pk]))
-        html = '<input type="checkbox" name="is_superuser" checked id="id_is_superuser">'
+        html = 'name="is_superuser"'
```
nemesisdesign: why did you have to change this?

openwisp_users/admin.py
```diff
@@ -139,10 +178,22 @@ def has_change_permission(self, request, obj=None):
...
         if not request.user.is_superuser:
-            qs = qs.filter(is_superuser=False)
+            org_ids = OrganizationUser.objects.only("organization_id") \
+                                              .values_list("organization_id") \
+                                              .filter(user=request.user,
+                                                      is_admin=True)
+            user_ids = OrganizationUser.objects.only("user_id") \
+                                               .values_list("user_id") \
+                                               .filter(organization__in=org_ids)
+            qs = User.objects.filter(pk__in=user_ids, is_superuser=False)
```
nemesisdesign: querying users by a list of IDs will be a problem in instances which use the freeradius moduile in which we expect organizations to have many users (potentially hundred of thousands). This code won't scale there because the full list of users for an organization will have to be fetched and built before being able to even query the DB and with a very large set of users this process can become slow, even cause timeouts of requests or halting the application server.

That's why last week I pushed this commit: https://github.com/openwisp/openwisp-users/commit/55ad1b3d9a199c33fa8e66517b51f0e363456561 which I think handled the problem well: for each organization the user is member of, a queryset is built to fetch users of that org except the requesting user. That queryset is not evaluated until executed so it will perform well. The code I wrote is more complex but sometimes that's a cost we can't avoid to sustain in order to write software which scales decently.

So what prompted you to replace that code with this code here?

nemesisdesign: PS: look at the current file you're editing, if single quotes `'` are used to delimit strings, please follow the same style. Let's not mix styles.

atb00ker: > So what prompted you to replace that code with this code here?

The problem with that code was that when a user has multiple common organisations with current operator, multiple records (for each organisation) are displayed on list and selecting any instance of such user throws an error.
However, i've been working on this problem and it looks like it can simply be solved by doing
`add_qs = org_user.organization.users.all().distinct()` -- i'll test it further and revert back to changes in that commit if everything works out, if not i'll discuss the problem here!

openwisp_users/tests/test_admin.py
atb00ker: True, it wouldn't make a difference, i'll change it back! :sweat_smile:

openwisp_users/admin.py
nemesisdesign: @atb00ker yes although this would fix #34 as well, which I think it's ok, we can do this

atb00ker: @nemesisdesign yes, thank you, #34 clears it.
I'll fix the behaviour in case:
if an operator `is_admin` in org1 and **not** `is_admin` in org2 and a member of default, org1 and org2 exist. Since, this member's record is visible to operator. The dropdown box doesn't contain the name of other organisations it looks like
this.

According to #34, i'll simply hide all other organisation's record and only org1 will be visible! :smile:

nemesisdesign: right.. this one we didn't consider.
This case right now shouldn't happen happen often, actually there's very little chance it will happen to any user right now, so I think we can create a new issue just for this case and go ahead, we'll live with the known bug for the moment while we think about a solution.

nemesisdesign: right.. this one we didn't consider.
This case right now shouldn't happen happen often, actually there's very little chance it will happen to any user right now, so I think we can create a new issue just for this case and go ahead, we'll live with the known bug for the moment while we think about a solution.

nemesisdesign: right.. this one we didn't consider.
This case right now shouldn't happen happen often, actually there's very little chance it will happen to any user right now, so I think we can create a new issue just for this case and go ahead, we'll live with the known bug for the moment while we think about a solution.

nemesisdesign: right.. this one we didn't consider.
This case right now shouldn't happen happen often, actually there's very little chance it will happen to any user right now, so I think we can create a new issue just for this case and go ahead, we'll live with the known bug for the moment while we think about a solution.

nemesisdesign: right.. this one we didn't consider.
This case right now shouldn't happen happen often, actually there's very little chance it will happen to any user right now, so I think we can create a new issue just for this case and go ahead, we'll live with the known bug for the moment while we think about a solution.

nemesisdesign: right.. this one we didn't consider.
This case right now shouldn't happen happen often, actually there's very little chance it will happen to any user right now, so I think we can create a new issue just for this case and go ahead, we'll live with the known bug for the moment while we think about a solution.

nemesisdesign: right.. this one we didn't consider.
This case right now shouldn't happen happen often, actually there's very little chance it will happen to any user right now, so I think we can create a new issue just for this case and go ahead, we'll live with the known bug for the moment while we think about a solution.

nemesisdesign: right.. this one we didn't consider.
This case right now shouldn't happen happen often, actually there's very little chance it will happen to any user right now, so I think we can create a new issue just for this case and go ahead, we'll live with the known bug for the moment while we think about a solution.

nemesisdesign: right.. this one we didn't consider.
This case right now shouldn't happen happen often, actually there's very little chance it will happen to any user right now, so I think we can create a new issue just for this case and go ahead, we'll live with the known bug for the moment while we think about a solution.

atb00ker: This bug will always be a problem for the case of #34 (because the operator himself would see similar results for his own record). Hence, i think #34 is more relevant for this discussion.
I have not done any action relating to solving #34 here, i think it'll need more changes (and this discussion), doing them here feels like mixing things. :smile:

nemesisdesign:
@atb00ker does this fix also #34?

atb00ker:
No, i've kept #34 out of this one.

nemesisdesign:
@atb00ker next time when you update a PR could you please ping me with a comment so I don't loose you in my feed? :joy:

PR #40, opened 2018-10-22
[groups] Added default "operator" and "user manager" groups.

### LINKED ISSUE #31: [groups] Add default "operator" and "admin" group
We should add two default groups:

- "operator" will be the group with permissions to see the basic OpenWISP objects needed to manage an organization (although each default permission available in this group should be added by each openwisp2 module, for example, the controller should add to this group the permissions to edit devices and templates, the radius module should add to this group the permissions to edit the radius related objects)
- "user manager" will have permission to manage users of their organization (although work is neeed to implement this in #32 and #34)

Other notes:
- operators won't have user editing permissions
- user managers will have permission to add/edit users so they must be able to assign groups to new users
- so groups have to be editable also by staff users (operators won't be able to change their group because they don't have permissions to edit user details)
- the important thing is that user managers can't create or change groups they can only choose from a predefined list of groups

R9295:
Implements and closes #31


openwisp_users/migrations/__init__.py
```diff
@@ -17,8 +20,33 @@ def set_default_organization_uuid(apps, schema_editor):
...
+    # To populate all the permissions
+    for app_config in apps.get_app_configs():
+        app_config.models_module = True
+        create_permissions(app_config, apps=apps, verbosity=0)
+        app_config.models_module = None
+
+    operator = group.objects.filter(name='operator')
+    if operator.count() == 0:
+        operator = group.objects.create(name='operator')
+        operator.permissions.set([])
```
nemesisdesign: this line is not needed now, right?

openwisp_users/migrations/__init__.py
```diff
@@ -17,8 +20,33 @@ def set_default_organization_uuid(apps, schema_editor):
...
+
+    user_manager = group.objects.filter(name='user manager')
+    if user_manager.count() == 0:
+        user_manager = group.objects.create(name='user manager')
+        user_manager.permissions.set([
+            Permission.objects.get(codename='add_user').pk,
+            Permission.objects.get(codename='change_user').pk,
+            Permission.objects.get(codename='view_user').pk,
```
nemesisdesign: this permission is available only from django 2.1 onwards (finally!). In order to support older versions of django the simples thing you can do in this case is to build the list of permissions (except the view one) before the `set` call, then append the `view_user` permission within a `try, except DoesNotExist:` block.

nemesisdesign:
PS: also rebase on the current master when you can

PR #41, opened 2018-10-24
[ci] Fixing Travis Build

atb00ker:
Fixed: Travis builds fail due to flake8 error on setup.py.

### DIFF
--- a/runflake8
+++ b/runflake8
@@ -1,4 +1,4 @@
 #!/bin/bash
 set -e
 flake8 --max-line-length=110 \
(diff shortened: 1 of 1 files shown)


atb00ker:
@nemesisdesign notice [this job](https://travis-ci.org/openwisp/openwisp-users/jobs/445588154) fails because of flake8 error on setup.py
Even jobs on master branch that were passing before are failing on restart, [check here](https://travis-ci.org/openwisp/openwisp-users/jobs/444312213)

Issue #48, opened 2018-10-01
[admin] Show readonly id of organization


R9295:
**Reserved for GCI**

In order to authenticate with the API, you need to know the organization's id and it's token. The token is displayed, but the org id however, is not. Create a **read-only** field  in ```OrganizationAdmin``` that displays the org's id.

PR #50, opened 2018-11-04
[admin] Added read-only id of organization #48

OltarzewskiK:
GCI task: https://codein.withgoogle.com/dashboard/task-instances/6235966491590656/


OltarzewskiK:
I probably finded solution!

OltarzewskiK:
Works I have to only write tests now!

OltarzewskiK:
As I said before `short_description` doesn't change name of field in my and `django-netjsonconfig` case, but I found a way to do that by jquery which works @atb00ker `UUID` field is now on top!

openwisp_users/tests/test_admin.py
```diff
@@ -322,3 +322,10 @@ def test_organization_owner(self):
         self._create_org_owner()
         response = self.client.get(reverse('admin:openwisp_users_organizationowner_changelist'))
         self.assertContains(response, 'tester')
+
+    def test_organization_id(self):
```
nemesisdesign: rename to `test_organization_uuid_field`

openwisp_users/admin.py
```diff
@@ -260,6 +260,28 @@ class OwnerInline(BaseOwnerInline):
 class OrganizationAdmin(BaseOrganizationAdmin, BaseAdmin):
     view_on_site = False
     inlines = [OwnerInline]
+    fields = ['id', 'name', 'is_active', 'slug', 'description', 'email', 'url']
```
nemesisdesign: remove from this line

openwisp_users/admin.py
```diff
@@ -260,6 +260,28 @@ class OwnerInline(BaseOwnerInline):
...
+        if obj:
+            return fields
+        new_fields = fields[:]
+        for field in self.readonly_fields:
+            if field in new_fields:
+                new_fields.remove(field)
+        return new_fields
+
+    def get_fields(self, request, obj=None):
+        return self._get_fields(self.fields, request, obj)
```
nemesisdesign: until this line, and replace all this block with the following code, which you should read, understand and learn to replicate yourself without my help in the future:

```python
    def id_str(self, obj):
        return str(obj.pk)

    id_str.short_description = 'UUID'

    def get_fields(self, request, obj=None):
        fields = super(OrganizationAdmin, self).get_fields(request, obj)
        if obj:
            fields = fields[:]
            fields.insert(0, 'id_str')
        return fields
```

openwisp_users/static/openwisp-users/js/uuid.js
```diff
@@ -0,0 +1,12 @@
...
+  "use strict";
+  $('.field-id label').text('UUID:');
+  var container = $('.field-id .readonly'),
+    value = container.text();
+  container.html('<input readonly id="id_id" type="text" class="vTextField readonly" value="' + value + '">');
+  var id = $('#id_id');
+  id.click(function () {
+    $(this).select();
+  });
+});
```
nemesisdesign: replace this code with the following:

```javascript
django.jQuery(function($) {
    var container = $('.field-id_str .readonly').eq(0),
        value = container.text();
    container.html('<input readonly id="id_id" type="text" class="vTextField readonly" value="'+ value +'">');
    var id = $('#id_id');
    id.click(function(){
        $(this).select();
    });
});
```

OltarzewskiK: In `openwisp-users` is `jslint` which requires 2 space instead of tab

openwisp_users/tests/test_admin.py
OltarzewskiK: Done

openwisp_users/admin.py
OltarzewskiK: This code doesn't work when trying to change organization is giving error however in organization add works properly but `readonly_field` is visible (it shouldn't) and name of this field is `Id:` not `UUID`

openwisp_users/static/openwisp-users/js/uuid.js
OltarzewskiK: and `jslint` requires to add `use strict` otherwise travis will return error

OltarzewskiK: And I think that there is no need to add `eq(0)` (if there is any reason for that please correct me)

openwisp_users/admin.py
OltarzewskiK:
```
def id_str(self, obj):
        return str(obj.pk)
```
and I think that `id` in this case isn't primary key because UUID has got letters too

openwisp_users/static/openwisp-users/js/uuid.js
nemesisdesign: improve it as needed, no problem

openwisp_users/admin.py
nemesisdesign: do that change and iterate to improve it / fix what's needed, I did that in 5 minutes so it's normal something may be missing, it's on your side to make it work, I'm showing you the right path to implement this feature with as less code as possible

OltarzewskiK: :+1:

OltarzewskiK:
I didn't find way to change name of field via django (even when I change value of `short_description` in `django-netjsonconfig` it doesn't change name of field). So I added it in by jquery. and I think that there is no need to add `eq(0)` here `var container = $('.field-id .readonly'),`

nemesisdesign:
don't change it with jQuery, just read my instructions and apply them as I indicated

OltarzewskiK:
@nemesisdesign `fields.insert(0, 'id_str')` won't work properly, because every time when website is refreshed `id_str` is added so after doing one refresh we have got 2 see `id` fields after 2 refreshes we see 3 etc.
So I think that removing fields is better than inserting (I did it in my commit)

OltarzewskiK:
I haven't changed `Id` to `UUID`

openwisp_users/admin.py
```diff
@@ -261,6 +261,20 @@ class OrganizationAdmin(BaseOrganizationAdmin, BaseAdmin):
...
 
+    readonly_fields = ['id']
+
+    def get_fields(self, request, obj=None):
+        fields = super(OrganizationAdmin, self).get_fields(request, obj)
+        fields = fields[:]
+        fields.remove('id')
+        if obj:
+            fields.insert(0, 'id')
+        return fields
```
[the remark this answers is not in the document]
OltarzewskiK: Yes, but when try to add organization `UUID` is set to `None`

atb00ker:
> @atb00ker this kind of logic is now repeated in 3 different openwisp modules. Do you think it's possible to extract it into a reusable widget or something like that? If yes we can create some tasks for it.

@nemesisdesign sorry, i lost this notification mail with other mails.
Yes, considering that it might be required in future as well, i think we should move this code to `openwisp-utils`. I'll update you on it as soon as possible. :smile:

PR #63, opened 2018-12-09
[admin] Move custom login form from openwisp-controller #33

### LINKED ISSUE #33: [admin] Move custom login form from openwisp-controller
I think the [custom login form in openwisp-controller](https://github.com/openwisp/openwisp-controller/blob/master/openwisp_controller/config/templates/admin/login.html) which allows to reset the password makes a lot more sense to be moved in this package since other modules may use openwisp-users and need the same feature but they may not need openwisp-controller.

ppabcd:
Fixes #33


ppabcd:
all done @nemesisdesign

PR #65, opened 2018-12-30
removed the query_set() function from UserAdmin and moved logic to multitenancy

### LINKED ISSUE #34: [admin] Only organizationuser objects which have admin=True should be able to manage their organization in the admin

shashwat1002:
But, I have serious concerns regarding the `runtests.py`, even after adding the new test function, number of tests didn't change, and to test the problem with test functions further, I deliberately did things to fail test like putting a particular item in both `visible` and `hidden` arguments of the `_test_multitenant_admin` function. I'll open an  issue for the same. 

closes #34

### DIFF
--- a/openwisp_users/multitenancy.py
+++ b/openwisp_users/multitenancy.py
@@ -2,6 +2,30 @@ from django.contrib import admin
 from django.db.models import Q
 from django.utils.translation import ugettext_lazy as _
 
+from .models import OrganizationUser, User
+
+
+def multitenant_behaviour_for_user_admin(self, request):
+    """
+    if operator is logged in - show only users
+    from same organization and hide superusers
+    if superuser is logged in - show all users
+    """
+    if not request.user.is_superuser:
+        user = request.user
+        org_users = OrganizationUser.objects.filter(user=user) \
+                                            .select_related('organization')
+        qs = User.objects.none()
+        for org_user in org_users:
+            if org_user.is_admin:
+                qs = qs | org_user.organization.users.all().distinct()
+        # hide superusers from organization operators
+        # so they can't edit nor delete them
+        qs = qs.filter(is_superuser=False)
+    else:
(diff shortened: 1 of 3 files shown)


shashwat1002:
Thanks! 😄  I feel stupid that I couldn't see this :33

shashwat1002:
Done @atb00ker 
@nemesisdesign

shashwat1002:
Done  @atb00ker

openwisp_users/multitenancy.py
```diff
@@ -2,6 +2,30 @@
 from django.db.models import Q
 from django.utils.translation import ugettext_lazy as _
 
+from .models import OrganizationUser, User
+
+
+def multitenant_behaviour_for_user_admin(self, request):
```
nemesisdesign: seems this function is only used in `MultitenantAdminMixin` right?
In this case I think it makes more sense to have this as a method of `MultitenantAdminMixin` instead of a stand-alone function. Can you do this change please?

shashwat1002: I think I actually wanted to use it as a static function somewhere, using the decorator, I don't remember now. Anyway, I just changed it.

shashwat1002:
Done @nemesisdesign

nemesisdesign:
Thanks @shashwat1002

PR #66, opened 2019-02-10
added redirect_view to urls, to directly land on '/admin' page.

vedangj044:
I think this is a small enhancement, as we would be redirected directly to, `admin` page, after entering `runserver` on console.


NoumbissiValere:
@vedangj044 one of the reasons why the test fails is because your didn't follow the commit style.
Please can you read this?
http://openwisp.io/docs/developer/contributing.html#commit-message-style-guidelines.
It will guide you on how to commit properly and i think travis will pass if that is ok

PR #70, opened 2019-02-23
[models] Add possibility to validate inverse relations #4

### LINKED ISSUE #4: Add possibility to validate inverse relations
Right now, we have an utility method called `_validate_org_relation`, which when added correctly used verifies that the organization of a related entity matches the current entity, for example:

**Ensuring a certificate is associated to the same organization of the related certification authority.**

This is cool, but then if we try to change organization on the Certification authority the system will allow us to do so, breaking all the certificates that are related to it because now they will be associated to a different organization than its Certification Authority.

**Note**: changing the organization on a CA doesn't really make any sense, it's just an example to explain the issue.

So we need to add a mechanism which performs this validation efficiently on inverse relations.

It could work like this:

- if the entity is new, do not perform the validation because there will be no related entities to it yet.
- if the entity is not new and the organization is being changed, count the objects that are related to it which have the old organization, if the number is higher than zero, fail the validation explaining that the organization cannot be changed because n number of related objects as still related to it

A test for this new validation method should be added too.

NoumbissiValere:
Added a reverse validation method which ensures that before the
organization of an object is changed, no object is related to it, else a
validation error is raised.
The method receives as arguments the models which are related to the
current entity and count for that the possible relations to this entity
with this particular organization.

Fixes #4

### DIFF
--- a/openwisp_users/mixins.py
+++ b/openwisp_users/mixins.py
@@ -27,6 +27,18 @@ class ValidateOrgMixin(object):
                                      related_object_label=rel._meta.verbose_name)
             raise ValidationError({field_error: message})
 
+    def _validate_org_reverse_relation(self, rel, field_error='organization'):
+        try:
+            old_org = self.__class__.objects.get(pk=self.pk).organization
+            if old_org != self.organization:
+                count = eval('self.__class__.objects.get(pk=self.pk).{0}_set.count()'.format(rel))
+                if count > 0:
+                    message = _('organization cannot be changed because {0}'
+                                ' objects as still related to it'.format(count))
+                    raise ValidationError({field_error: message})
+        except self.__class__.DoesNotExist:
+            return
+
 
 class OrgMixin(ValidateOrgMixin, models.Model):
     """
--- a/tests/testapp/tests.py
+++ b/tests/testapp/tests.py
@@ -40,6 +40,16 @@ class TestIntegration(TestOrganizationMixin, TestCase):
         with self.assertRaises(ValidationError):
             c.full_clean()
 
+    def test_validate_reverse_org_relation(self):
+        org1 = self._create_org(name='org1')
+        org2 = self._create_org(name='org2')
+        t = Template.objects.create(name='test-t', organization=org1)
+        Config.objects.create(name='test-c1', template=t, organization_id=str(org1.pk))
+        with self.assertRaises(ValidationError):
+            t.organization = org2
+            t.full_clean()
+            t.save()
+
     def test_resolve_account_URLs(self):
         resolver = resolve('/accounts/login/')
         self.assertEqual(resolver.view_name, 'account_login')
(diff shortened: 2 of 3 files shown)


NoumbissiValere:
@atb00ker please can you look at this when you have time? so that i can perform changes if needed.
Thanks

openwisp_users/mixins.py
```diff
@@ -27,6 +27,18 @@ def _validate_org_relation(self, rel, field_error='organization'):
                                      related_object_label=rel._meta.verbose_name)
             raise ValidationError({field_error: message})
 
+    def _validate_org_reverse_relation(self, rel, field_error='organization'):
+        try:
+            old_org = self.__class__.objects.get(pk=self.pk).organization
+            if old_org != self.organization:
+                count = eval('self.__class__.objects.get(pk=self.pk).{0}_set.count()'.format(rel))
```
nemesisdesign: this can be achieved without using eval, eg: 

```python
count = getattr(old_self, '{0}_set'.format(rel)).count()
```

openwisp_users/mixins.py
```diff
@@ -27,6 +27,18 @@ def _validate_org_relation(self, rel, field_error='organization'):
...
+                if count > 0:
```
nemesisdesign: `if count:` is enough

tests/testapp/tests.py
```diff
@@ -40,6 +40,16 @@ def test_validate_org_relation_error(self):
         with self.assertRaises(ValidationError):
             c.full_clean()
 
+    def test_validate_reverse_org_relation(self):
+        org1 = self._create_org(name='org1')
+        org2 = self._create_org(name='org2')
+        t = Template.objects.create(name='test-t', organization=org1)
+        Config.objects.create(name='test-c1', template=t, organization_id=str(org1.pk))
```
nemesisdesign: use `organization=org1` instead of `organization_id=str(org1.pk)` (for consistency with the previous line, which is also easier to read).

openwisp_users/mixins.py
```diff
@@ -27,6 +27,18 @@ def _validate_org_relation(self, rel, field_error='organization'):
                                      related_object_label=rel._meta.verbose_name)
             raise ValidationError({field_error: message})
 
+    def _validate_org_reverse_relation(self, rel, field_error='organization'):
+        try:
```
nemesisdesign: `old_self = self.__class__.objects.get(pk=self.pk)`

openwisp_users/mixins.py
```diff
@@ -27,6 +27,18 @@ def _validate_org_relation(self, rel, field_error='organization'):
...
+            old_org = self.__class__.objects.get(pk=self.pk).organization
```
nemesisdesign: `old_org = old_self.organization`

openwisp_users/mixins.py
```diff
@@ -27,6 +27,18 @@ def _validate_org_relation(self, rel, field_error='organization'):
...
+            if old_org != self.organization:
+                count = eval('self.__class__.objects.get(pk=self.pk).{0}_set.count()'.format(rel))
+                if count > 0:
+                    message = _('organization cannot be changed because {0}'
+                                ' objects as still related to it'.format(count))
```
nemesisdesign: here we need to add the type of object that is still related, using `rel`

PR #71, opened 2019-02-28
[models] Add unique constraint on user.email #28

### LINKED ISSUE #28: [models] Ensure emails are unique
~~I assumed that the default django user model had a unique constraint on the email field, but I was wrong.~~

~~We need to add this.~~

After further testing and thought I concluded that we can't do this or we won't be able to generate freeradius users with no email address.

We can't add a database constraint but we can add application validation logic which checks that if a user has an email set, no other user except him has the same email, otherwise raise a `ValidationError`.
Actually, if the blank email is set to `None`, the unique constraint error shouldn't be triggered so that may be feasible.

NoumbissiValere:
Added a clean method in the users model which validates if the
email already exist in the database or not.
A test case was added to test for this functionality

Fixes #28


NoumbissiValere:
@atb00ker Please can you review this? Thanks

NoumbissiValere:
> Oops, I am unable to make a user with no email ID.
> Did you check that? 
> 
> I think the issue wants us to do the following:
> If an email address is entered - check if it's unique, else, raise error.
> Let the user be created if no email is entered. i.e skip check.

Ok @atb00ker . I will add this check and send. Thanks for reviewing. :smiley:

NoumbissiValere:
I have tried again to increase the coverage but still doesn't go up :disappointed:

openwisp_users/admin.py
```diff
@@ -204,7 +184,7 @@ def get_inline_instances(self, request, obj=None):
         if request:
             if hasattr(inline, '_has_add_permission'):
                 has_add_perm = inline._has_add_permission(request, obj)
-            else:
+            if hasattr(inline, 'has_add_permission'):
```
atb00ker: Is this possible?
```
if hasattr(inline, '_has_add_permission') or hasattr(inline, 'has_add_permission'):
```

atb00ker:
> I have tried again to increase the coverage but still doesn't go up

The coverage has gone down because a lot of lines have been reduced, although, i think since this coverage improvement is not related to this pull request and this can be merged, if you want you can checkout the lines that need testing here: https://coveralls.io/builds/22232224/source?filename=openwisp_users/multitenancy.py

openwisp_users/admin.py
NoumbissiValere: I tried doing that but i noticed inline can have `_has_add_permission` and/or `has_add_permission` in some situations. thus two conditions will be needed to make `inline.has_add_permission` and `inline._has_add_permission` to dynamically take care of the situations.
I Equally tried doing this;
```
if request:
    has_add_perm=inline.has_add_permission(request)
    if hasattr(inline, '_has_add_permission'):
        has_add_perm = inline._has_add_permission(request, obj)
```
which will override the value of `has_add_perm` if `_has_add_permission` exist since i noticed priority was given to the `_has_add_permission` attribute.
But that didn't increase the coverage.

NoumbissiValere:
> > I have tried again to increase the coverage but still doesn't go up
> 
> The coverage has gone down because a lot of lines have been reduced, although, i think since this coverage improvement is not related to this pull request and this can be merged, if you want you can checkout the lines that need testing here: https://coveralls.io/builds/22232224/source?filename=openwisp_users/multitenancy.py

Ok

openwisp_users/admin.py
nemesisdesign: I do not understand the nature of the change on this line, we should check the django documentation or source to understand if we can do this, @NoumbissiValere is there a chance you did find out?

As far as I know, the old code was working fine and I would not change it unless it really makes sense to do it.
Having stable and predictable code has the priority over test coverage.

openwisp_users/models.py
```diff
@@ -59,6 +60,14 @@ def organizations_pk(self):
                     .values_list('organization_id')
         return qs
 
+    def clean(self):
+        if self.email:
```
nemesisdesign: could you add a comment that recaps what I wrote in the issue description? I mean to write the reason for which we are doing application level validation and not using the DB constraint, it's explained in the description of #28

openwisp_users/models.py
```diff
@@ -59,6 +60,14 @@ def organizations_pk(self):
...
+            try:
+                User.objects.exclude(pk=self.pk).get(email=self.email)
```
nemesisdesign: use `User.objects.exclude(pk=self.pk).filter(email=self.email).exists()` which is going to be more efficient.

It could be:

```python
if User.objects.exclude(pk=self.pk).filter(email=self.email).exists():
    raise ValidationError({'email': 'A user with this email already exist'})
```

openwisp_users/models.py
```diff
@@ -59,6 +60,14 @@ def organizations_pk(self):
...
+                raise ValidationError({'email': 'A user with this email already exist'})
+            except User.DoesNotExist:
+                pass
+
```
nemesisdesign: could you also add a new index to make sure that query runs fast?

```
index_together = (
    ('pk', 'email')
)
```

Read more about https://docs.djangoproject.com/en/2.2/ref/models/options/#index-together

openwisp_users/tests/test_admin.py
```diff
@@ -90,9 +90,9 @@ def test_admin_change_user_email_empty(self):
         })
         response = self.client.post(reverse('admin:openwisp_users_user_change', args=[admin.pk]), params)
         queryset = User.objects.filter(username='testchange')
-        self.assertEqual(queryset.count(), 0)
+        self.assertEqual(queryset.count(), 1)
         self.assertEqual(len(mail.outbox), 0)
-        self.assertContains(response, 'errors field-email')
+        self.assertEqual(response.status_code, 302)
```
nemesisdesign: this is not correct, we still want the email field to be required.
Users cannot create users without the email field. Only the system itself can (eg: the django-freeradius module).

openwisp_users/tests/test_models.py
```diff
@@ -18,6 +19,23 @@ def test_create_superuser_email_empty(self):
...
+                'username': 'user2',
+                'email': 'same@gmail.com',
+                'password': 'pass1'
+            }
+            u = User(**options)
+            u.full_clean()
+            u.save()
+
+    def test_create_user_without_email(self):
+        User.objects.create(username='testuser', password='test1', email='')
```
nemesisdesign: do not use `create()`, because it does not run the validation, instantiate the user and then validate it with `full_clean()`

NoumbissiValere:
Thanks @nemesisdesign for taking out your precious time to review this PR. I will look into it and make the changes requested

NoumbissiValere:
@nemesisdesign Please can you review the changes i have made? Thanks

openwisp_users/migrations/0001_initial.py
```diff
@@ -144,4 +144,8 @@ class Migration(migrations.Migration):
             name='organizationuser',
             unique_together=set([('user', 'organization')]),
         ),
+        migrations.AlterIndexTogether(
```
nemesisdesign: a new migration should be created for this, otherwise existing instances will not get the new index

openwisp_users/models.py
```diff
@@ -47,6 +48,9 @@ class User(AbstractUser):
 
     objects = UserManager()
 
+    class Meta(AbstractUser.Meta):
+        index_together = ['id', 'email']
```
nemesisdesign: can we change it to use tuple please? `('id', 'email')`

openwisp_users/models.py
```diff
@@ -59,6 +63,19 @@ def organizations_pk(self):
...
+        """
+        If a user has an email set, validate the unique constraint
+        else don't validate this constraint.
+        If no email is set, raise a validation error
+        """
+        if self.email:
+            if User.objects.exclude(pk=self.pk).filter(email=self.email).exists():
+                raise ValidationError({'email': 'A user with this email already exist'})
+        else:
+            raise ValidationError({'email': 'Please this field is required'})
```
nemesisdesign: we cannot do this or modules that need to create users without email won't work, this can happen sometimes, read here: https://django-freeradius.readthedocs.io/en/latest/general/generating_users.html

This is one of those few times in which we need to use the form level validation instead of the model level validation: we allow the email to be blank on the model, so we leave that option open for any feature that may need it, but we use the form level validation in the admin to enforce the email to be present when admins use the web interface to add users

openwisp_users/admin.py
```diff
@@ -79,24 +76,7 @@ def get_extra(self, request, obj=None, **kwargs):
         return 0
 
 
-class EmailRequiredMixin(forms.ModelForm):
```
nemesisdesign: so this needs to be brought back

NoumbissiValere:
@nemesisdesign travis flag the new migration file as conflicting with `0005_user_phone_number` when i rename it to descriptive name. and running the command `python manage.py makemigrations --merge` on my local repo produces no change.

nemesisdesign:
Great work @NoumbissiValere :+1:

PR #74, opened 2019-05-02
[admin] Server Error (500). openwisp#72

NoumbissiValere:
log the exception sendmail raised during
user creation with invalid email so that
users are created even with invalid email.
Added a mock test to mimic the exception
raised during user creation.
Added mock to requirements-test.txt

Fixes openwisp#72

### DIFF
--- a/openwisp_users/admin.py
+++ b/openwisp_users/admin.py
@@ -1,3 +1,4 @@
+import logging
 from copy import deepcopy
 
 from allauth.account.models import EmailAddress
@@ -21,6 +22,8 @@ from .base import BaseAdmin
 from .models import (Group, Organization, OrganizationOwner, OrganizationUser,
                      User)
 from .multitenancy import MultitenantAdminMixin
+
+logger = logging.getLogger(__name__)
 
 
 class EmailAddressInline(admin.StackedInline):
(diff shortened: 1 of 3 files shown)


NoumbissiValere:
I replicated the issue on my `openwisp_users`  by the setting the `DEFAULT_FROM_EMAIL` in the `settings.py` as it was in `vagrant-openwisp2` VM. the tests passed when everything was set back to normal and i equally installed and tested it on `vagrant-openwisp2` VM with the normal settings and it worked well.
I don't know much about issue #132 but when this patch was applied on my vagrant VM, i didn't get the database locked problem when i changed the admin email to an invalid one. so maybe this issue is related to that and maybe this patch can fix it (haven't tested that)

@nemesisdesign please can you review this when you have time? 
thanks.

openwisp_users/admin.py
```diff
@@ -217,11 +221,14 @@ def save_model(self, request, obj, form, change):
...
-                                           email=obj.email,
-                                           confirm=True,
-                                           signup=True)
+            try:
+                EmailAddress.objects.add_email(request,
+                                               user=obj,
+                                               email=obj.email,
+                                               confirm=True,
+                                               signup=True)
+            except smtplib.SMTPSenderRefused as e:
```
nemesisdesign: just catch `Exception`

openwisp_users/admin.py
```diff
@@ -217,11 +221,14 @@ def save_model(self, request, obj, form, change):
...
+                logger.exception(e)
```
nemesisdesign: `logger.exception()` will automatically log the stack trace of the exception being captured, the argument passed to the method is used to add additional information, so passing it `e` won't be very useful.

Change this line to:

```python 
logger.exception(
    'Got exception {} while sending '
    'verification email to user {}, email {}'.format(
        type(e), obj.username, obj.email
    )
)
```

openwisp_users/tests/test_admin.py
```diff
@@ -333,6 +336,20 @@ def test_organization_uuid_field(self):
...
+        params = dict(username='testmail',
+                      email='test@invalid.com',
+                      password1='tester',
+                      password2='tester')
+        params.update(self.add_user_inline_params)
+        with mock.patch('allauth.account.models.EmailAddress.objects.add_email') as mocked:
+            mocked.side_effect = smtplib.SMTPSenderRefused(501, '5.1.7 Bad sender address syntax',
+                                                           'test_name@test_domain')
+            self.client.post(reverse('admin:openwisp_users_user_add'), params)
+            mocked.assert_called_once()
```
nemesisdesign: looks good :+1:

NoumbissiValere:
@nemesisdesign the changes have been made. Please can you review again? 
Thanks

Issue #75, opened 2019-05-16
[API] Add an API for CRUD of users and organization


NoumbissiValere:
- Add an API which will enable users to signup using a social account. when a user successfully signs up, they should be directed to create an organization before they can be allowed to create templates
- Add an API which will enable users to list, edit, delete owned organizations and view the details of a particular organization
- Add an API which will enable users to retrieve and edit their user profile

These APIs should be disabled by default and only enabled at the level of the template library.

PR #81, opened 2019-07-11
[api] Added API for CRUD of organization #75

NoumbissiValere:


NoumbissiValere:
This functionality have been moved to the library backend as asked. Thus, i think it should be closed

PR #83, opened 2019-12-03
[Travis] Fixed Warnings

Weirdo914:
[warn] on root: deprecated key: "sudo" (The key `sudo` has no effect anymore.)

[info] on root: the key matrix is an alias for jobs, using jobs

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: python
-sudo: false
+os: linux
 cache: pip
 
 python:


marfgold1:
Whoops i didn't realize you already working at it haha

.travis.yml
```diff
@@ -1,5 +1,5 @@
 language: python
-sudo: false
+os: linux
```
atb00ker: I checked this is only appearing as info not as warning so I think we can reverse it for now.
(The thing is that these things keep changing, so it's not worth our time to keep the info part changing, only when it appears as a warning that we need to do something about it! smile )

PR #85, opened 2019-12-04
[Travis] Fix warnings 

hax0kartik:


.travis.yml
```diff
@@ -1,5 +1,5 @@
 language: python
-sudo: false
+os: linux
```
atb00ker: I checked this is only appearing as info not as warning so I think we can reverse it for now.
(The thing is that these things keep changing, so it's not worth our time to keep the info part changing, only when it appears as a warning that we need to do something about it! smile )

.travis.yml
```diff
@@ -11,7 +11,7 @@ env:
   - DJANGO="django>=2.0,<2.1"
   - DJANGO="django>=2.1,<2.2"
 
-matrix:
+jobs:
```
atb00ker: I checked this is only appearing as info not as warning so I think we can reverse it for now.
(The thing is that these things keep changing, so it's not worth our time to keep the info part changing, only when it appears as a warning that we need to do something about it! smile )

hax0kartik: Right, I thought since I was at it, I might as well fix these too. :smile:
Pushed a commit which reverts this, thanks for the review!

PR #86, opened 2019-12-11
[ux] Add organization owner validation #76

### LINKED ISSUE #76: [ux] Server error 500 received instead form indicating error
When adding a new organization and selecting an organization user from a different organization server error 500 is recevied.

Expected: Form indicating "please select user from same organization"

pawelplsi:
Fixes #76

### DIFF
--- a/openwisp_users/models.py
+++ b/openwisp_users/models.py
@@ -5,6 +5,7 @@ from allauth.account.models import EmailAddress
 from django.contrib.auth.models import AbstractUser
 from django.contrib.auth.models import Group as BaseGroup
 from django.contrib.auth.models import UserManager as BaseUserManager
+from django.core.exceptions import ValidationError
 from django.db import models
 from django.utils.encoding import python_2_unicode_compatible
 from django.utils.functional import cached_property
@@ -118,3 +119,8 @@ class OrganizationOwner(AbstractOrganizationOwner):
     OpenWISP OrganizationOwner model
     """
     id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
+
+    def clean(self):
+        if self.organization_user.organization.pk != self.organization.pk:
+            raise ValidationError({
+                'organization_user': _('The selected user is not member of this organization.')})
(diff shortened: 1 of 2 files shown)


openwisp_users/tests/test_models.py
```diff
@@ -61,6 +61,18 @@ def test_organization_repr(self):
...
+        user = self._create_user(username='user1', email='abc@example.com')
+        org1 = self._create_org(name='org1')
+        org2 = self._create_org(name='org2')
+        org_user = self._create_org_user(organization=org1, user=user)
+        org_owner = self._create_org_owner()
+        org_owner.organization = org2
+        org_owner.organization_user = org_user
+        with self.assertRaises(ValidationError):
+            org_owner.full_clean()
+            org_owner.save()
```
atb00ker: I think the error would be saved at "org_owner.full_clean()", right?, so please remove `org_owner.save()` from the assertRasises scope.

Why? Because we want only the line that can raise the error and not anyother line, so that if there is any other error in the `org_owner.save()` line, it's not hidden from us! :smile:

pawelplsi: Yes, but it seems to be like that in all the other tests, why it is so then?

nemesisdesign: I found only one and I removed it: https://github.com/openwisp/openwisp-users/commit/4f64f7cd76987cbeff7115aae7e3187823a84ce2

Do you see other instances of this?

@atb00ker is right.

openwisp_users/models.py
```diff
@@ -118,3 +119,7 @@ class OrganizationOwner(AbstractOrganizationOwner):
     OpenWISP OrganizationOwner model
     """
     id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
+
+    def clean(self):
+        if self.organization_user.organization.pk != self.organization.pk:
+            raise ValidationError({'organization_user': 'Please select user from the same organization'})
```
nemesisdesign: change this to:

```python
raise ValidationError({
    'organization_user': _('The selected user is not member of this organization.)'
})```

pawelplsi: Done

PR #88, opened 2019-12-13
[requirements] Update dependencies & drop python 2 support #82

### LINKED ISSUE #82: [requirements] Update dependencies & drop python 2 support
Python2 is very close to the end of it's lifecycle, it's time we drop the support for Python 2.

To do that,
1. We need to update all the dependencies to python3.
2. Remove code that exists for python 2 support.
3. Remove python2 from README & any documentation.

Weirdo914:
Upadated dependencies and dropped support for python 2
Fixes #82


.travis.yml
```diff
@@ -2,20 +2,12 @@ language: python
 cache: pip
 
 python:
-  - "3.5"
-  - "2.7"
+  - "3.7"
 
 env:
-  - DJANGO="django>=1.11,<1.12"
```
Weirdo914: Django-allauth does not support django 1.11 so I removed this,

Weirdo914: Django-allauth does not support Django 1.11 so removed this.

requirements-test.txt
```diff
@@ -2,5 +2,4 @@ coverage
 coveralls
 isort
 flake8
-openwisp-utils[qa]>=0.2.1
-mock>=2.0.0
+openwisp-utils[qa]>=0.2.1
```
nemesisdesign: set `openwisp-utils[qa]>=0.3.0`

requirements.txt
```diff
@@ -1,6 +1,6 @@
 django-organizations>=1.0.0,<1.2.0
-django-extensions>=1.7.5,<2.2
-django-allauth>=0.35.0,<0.39.0
-django-phonenumber-field>=2.3.1,<2.4.0
-phonenumbers>=8.10.8,<8.11.0
+django-extensions>=1.7.5,<2.3
+django-allauth>=0.40
```
nemesisdesign: always put an upper limit to the version to avoid future versions of the dependency to break the build

requirements.txt
```diff
@@ -1,6 +1,6 @@
...
+django-phonenumber-field>=4.0
```
nemesisdesign: always put an upper limit to the version to avoid future versions of the dependency to break the build

Weirdo914:
Closed as #95 handles it well

PR #102, opened 2020-03-08
[admin] Add possibility to deactivate users in batch operation

### LINKED ISSUE #98: [admin] Add possibility to deactivate users in batch operation
Use case:

Go to list of users, select different users, make them inactive with one batch operation.

nepython:
Closes #98

### DIFF
--- a/openwisp_users/admin.py
+++ b/openwisp_users/admin.py
@@ -4,12 +4,14 @@ from copy import deepcopy
 from allauth.account.models import EmailAddress
 from django import forms
 from django.apps import apps
-from django.contrib import admin
+from django.contrib import admin, messages
+from django.contrib.admin.utils import model_ngettext
 from django.contrib.auth.admin import GroupAdmin as BaseGroupAdmin
 from django.contrib.auth.admin import UserAdmin as BaseUserAdmin
 from django.contrib.auth.forms import UserChangeForm as BaseUserChangeForm
 from django.contrib.auth.forms import UserCreationForm as BaseUserCreationForm
 from django.forms.models import BaseInlineFormSet
+from django.template.response import TemplateResponse
 from django.utils.translation import ugettext_lazy as _
 from openwisp_utils.admin import UUIDAdmin
 from organizations.base_admin import (BaseOrganizationAdmin,
@@ -142,6 +144,46 @@ class UserAdmin(MultitenantAdminMixin, BaseUserAdmin, BaseAdmin):
                     'last_login']
     inlines = [EmailAddressInline, OrganizationUserInline]
     save_on_top = True
+    actions = ['make_inactive', 'make_active']
+
+    def require_confirmation(func):
+        """
+        Decorator to lead to a confirmation page.
+        This has been used rather than simply adding the same lines in action functions
+        inorder to avoid repetition of the same lines in the two added actions and more actions
+        incase they are added in future.
+        """
+        def wrapper(modeladmin, request, queryset):
+            opts = modeladmin.model._meta
+            if request.POST.get('confirmation') is None:
+                request.current_app = modeladmin.admin_site.name
+                context = {'action': request.POST['action'],
+                           'queryset': queryset,
+                           'opts': opts}
+                return TemplateResponse(request, 'admin/action_confirmation.html', context)
+            return func(modeladmin, request, queryset)
+        wrapper.__name__ = func.__name__
+        return wrapper
+
+    @require_confirmation
+    def make_inactive(self, request, queryset):
+        queryset.update(is_active=False)
+        count = queryset.count()
+        if count:
+            self.message_user(request,
+                              _(f'Successfully made {count} {model_ngettext(self.opts, count)} inactive.'),
+                              messages.SUCCESS)
+    make_inactive.short_description = _('Flag selected users as inactive')
+
+    @require_confirmation
+    def make_active(self, request, queryset):
+        queryset.update(is_active=True)
+        count = queryset.count()
+        if count:
+            self.message_user(request,
+                              _(f'Successfully made {count} {model_ngettext(self.opts, count)} active.'),
+                              messages.SUCCESS)
+    make_active.short_description = _('Flag selected users as active')
 
     def get_list_display(self, request):
         """
--- a/openwisp_users/tests/test_admin.py
+++ b/openwisp_users/tests/test_admin.py
@@ -493,6 +493,57 @@ class TestUsersAdmin(TestOrganizationMixin, TestCase):
         html = '<input type="text" name="name" value="default"'
         self.assertNotContains(response, html)
 
+    def test_action_active(self):
+        user = User.objects.create(username='openwisp',
+                                   password='test',
+                                   email='openwisp@test.com',
+                                   is_active=False)
+        path = reverse('admin:openwisp_users_user_changelist')
+        self.client.force_login(self._get_admin())
+        post_data = {
+            '_selected_action': [user.pk],
+            'action': 'make_active',
+            'csrfmiddlewaretoken': 'test',
+            'confirmation': 'Confirm',
+            }
+        response = self.client.post(path, post_data, follow=True)
+        self.assertEqual(response.status_code, 200)
+
+    def test_action_inactive(self):
+        user = User.objects.create(username='openwisp',
+                                   password='test',
+                                   email='openwisp@test.com',
+                                   is_active=True)
+        path = reverse('admin:openwisp_users_user_changelist')
+        self.client.force_login(self._get_admin())
+        post_data = {
+            '_selected_action': [user.pk],
+            'action': 'make_inactive',
+            'csrfmiddlewaretoken': 'test',
+            'confirmation': 'Confirm',
+            }
+        response = self.client.post(path, post_data, follow=True)
+        user.refresh_from_db()
+        self.assertFalse(user.is_active)
+        self.assertEqual(response.status_code, 200)
+
+    def test_action_confirmation_page(self):
+        user = User.objects.create(username='openwisp',
+                                   password='test',
+                                   email='openwisp@test.com',
+                                   is_active=True)
+        path = reverse('admin:openwisp_users_user_changelist')
+        self.client.force_login(self._get_admin())
+        post_data = {
+            '_selected_action': [user.pk],
+            'action': 'make_active',
+            'csrfmiddlewaretoken': 'test',
+            }
+        response = self.client.post(path, post_data, follow=True)
+        user.refresh_from_db()
+        self.assertTrue(user.is_active)
+        self.assertEqual(response.status_code, 200)
+
     @patch('sys.stdout', devnull)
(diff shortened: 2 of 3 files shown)


nepython:
@nemesisdesign, can you please review this. Also, though the issue #102  mentioned about allowing to deactivate users in a batch operation, I found it appropriate to add one more to make them active again(I think it should not be a problem):grinning:

openwisp_users/tests/test_admin.py
```diff
@@ -493,6 +499,18 @@ def test_organization_uuid_field(self):
         html = '<input type="text" name="name" value="default"'
         self.assertNotContains(response, html)
 
+    def test_admin_actions(self):
+        request = MockRequest()
```
atb00ker: Can you please explain this, Why is this an object of `MockRequest` which is empty! :smile:

openwisp_users/admin.py
```diff
@@ -142,6 +142,15 @@ class UserAdmin(MultitenantAdminMixin, BaseUserAdmin, BaseAdmin):
                     'last_login']
     inlines = [EmailAddressInline, OrganizationUserInline]
     save_on_top = True
+    actions = ['make_inactive', 'make_active']
+
+    def make_inactive(self, request, queryset):
+        queryset.update(is_active=False)
+    make_inactive.short_description = 'make selected users inactive'
```
atb00ker: Can you rename description to "Deactivate account(s)". :smile:

openwisp_users/admin.py
```diff
@@ -142,6 +142,15 @@ class UserAdmin(MultitenantAdminMixin, BaseUserAdmin, BaseAdmin):
...
+
+    def make_active(self, request, queryset):
+        queryset.update(is_active=True)
+    make_active.short_description = 'make selected users active'
```
atb00ker: Can you rename description to "Activate account(s)". :smile:

nepython:
@atb00ker, thanks for reviewing it. I have implemented your suggested changes can you please review them.:grinning:

openwisp_users/admin.py
```diff
@@ -142,6 +142,15 @@ class UserAdmin(MultitenantAdminMixin, BaseUserAdmin, BaseAdmin):
                     'last_login']
     inlines = [EmailAddressInline, OrganizationUserInline]
     save_on_top = True
+    actions = ['make_inactive', 'make_active']
+
+    def make_inactive(self, request, queryset):
+        queryset.update(is_active=False)
+    make_inactive.short_description = 'Deactivate account(s)'
```
nemesisdesign: use `gettext_lazy` to make this translatable:

```python
make_inactive.short_description = _('Flag selected users as inactive')
```

Everytime you write anything that will be presented to the user, you must do this.

Note how I changed the string too, I want to avoid "Deactivate" because it's too similar to "Delete" which is the other admin action users may click by mistake on delete instead of deactivate.

I also removed `(s)` for consistency with `Delete selected users`.

openwisp_users/admin.py
```diff
@@ -142,6 +142,15 @@ class UserAdmin(MultitenantAdminMixin, BaseUserAdmin, BaseAdmin):
...
+
+    def make_active(self, request, queryset):
+        queryset.update(is_active=True)
+    make_active.short_description = 'Activate account(s)'
```
nemesisdesign: same here, use `_('Flag selected users as activte')`

nepython:
I have added the confirmation page and message
Screenshot from 2020-03-11 20-54-58
Screenshot from 2020-03-11 20-55-08
Though I still need to add test cases for the same, posting these just for an update:smile:

openwisp_users/admin.py
nepython: @nemesisdesign thanks for reviewing, though i had previously used a similar string
`make_inactive.short_description = 'make selected users inactive'` but was suggested to use this one so went for it:smile:. Also, I have implemented your suggested changes and attached pics for reference.

nemesisdesign: ok, well the good thing is that at least is not a lot of work to change it. @atb00ker keep in mind this for future refrence, always keep consistency with what's already present in the system

openwisp_users/admin.py
```diff
@@ -142,6 +144,42 @@ class UserAdmin(MultitenantAdminMixin, BaseUserAdmin, BaseAdmin):
...
+    def require_confirmation(func):
+        def wrapper(modeladmin, request, queryset):
+            opts = modeladmin.model._meta
+            if request.POST.get('confirmation') is None:
+                request.current_app = modeladmin.admin_site.name
+                context = {'action': request.POST['action'],
+                           'queryset': queryset,
+                           'opts': opts}
+                return TemplateResponse(request, 'admin/action_confirmation.html', context)
+
```
nemesisdesign: could you please remove this blank line to make the code more compact?

openwisp_users/admin.py
```diff
@@ -142,6 +144,42 @@ class UserAdmin(MultitenantAdminMixin, BaseUserAdmin, BaseAdmin):
                     'last_login']
     inlines = [EmailAddressInline, OrganizationUserInline]
     save_on_top = True
+    actions = ['make_inactive', 'make_active']
+
+    def require_confirmation(func):
```
nemesisdesign: nice solution!
Can you add a docstring indicating that this is a decorator function and what it does?
This will help future contributors and even ourselves.

openwisp_users/admin.py
```diff
@@ -142,6 +144,42 @@ class UserAdmin(MultitenantAdminMixin, BaseUserAdmin, BaseAdmin):
...
+
+            return func(modeladmin, request, queryset)
+
+        wrapper.__name__ = func.__name__
+        return wrapper
+
+    @require_confirmation
+    def make_inactive(self, request, queryset):
+        queryset.update(is_active=False)
+        n = queryset.count()
```
nemesisdesign: please change `n` to `count` which is more explicit

openwisp_users/admin.py
```diff
@@ -142,6 +144,42 @@ class UserAdmin(MultitenantAdminMixin, BaseUserAdmin, BaseAdmin):
...
+        if n:
+            self.message_user(request,
+                              _(f'Successfully made {n} {model_ngettext(self.opts, n)} inactive.'),
+                              messages.SUCCESS)
+    make_inactive.short_description = _('Flag selected users as inactive')
+
+    @require_confirmation
+    def make_active(self, request, queryset):
+        queryset.update(is_active=True)
+        n = queryset.count()
```
nemesisdesign: please change `n` to `count` which is more explicit

openwisp_users/tests/test_admin.py
```diff
@@ -493,6 +493,44 @@ def test_organization_uuid_field(self):
         html = '<input type="text" name="name" value="default"'
         self.assertNotContains(response, html)
 
+    def test__action_active(self):
```
nemesisdesign: there's a double underscore (`__`) in the test function name, can you remove one please?

openwisp_users/tests/test_admin.py
```diff
@@ -493,6 +493,44 @@ def test_organization_uuid_field(self):
...
+        user = User.objects.create(is_active=False)
```
nemesisdesign: please don't do it this way, creating a user like that (without any info set except `is_active`) generates an invalid user into the DB, writing tests on an invalid model instance is really fragile.

Use this method to create an user, https://github.com/openwisp/openwisp-users/blob/master/openwisp_users/tests/utils.py#L31 then flag it as inactive.

openwisp_users/tests/test_admin.py
```diff
@@ -493,6 +493,44 @@ def test_organization_uuid_field(self):
...
+        path = reverse('admin:openwisp_users_user_changelist')
+        self.client.force_login(self._get_admin())
+        post_data = {
+                '_selected_action': [user.pk],
+                'action': 'make_active',
+                'csrfmiddlewaretoken': 'test',
+                'confirmation': 'Confirm',
+            }
+        response = self.client.post(path, post_data, follow=True)
+        self.assertEqual(response.status_code, 200)
```
nemesisdesign: please ensure the user has been flagged as active after the operation:

```python
user.refresh_from_db()
self.assertTrue(user.is_active)
```

Without this final assertion, the whole test method is useless, because the feature can break and the test will pass anyway.

openwisp_users/tests/test_admin.py
```diff
@@ -493,6 +493,44 @@ def test_organization_uuid_field(self):
...
+
+    def test__action_inactive(self):
+        user = User.objects.create(is_active=True)
```
nemesisdesign: same concept here

openwisp_users/tests/test_admin.py
```diff
@@ -493,6 +493,44 @@ def test_organization_uuid_field(self):
...
+        path = reverse('admin:openwisp_users_user_changelist')
+        self.client.force_login(self._get_admin())
+        post_data = {
+                '_selected_action': [user.pk],
+                'action': 'make_inactive',
+                'csrfmiddlewaretoken': 'test',
+                'confirmation': 'Confirm',
+            }
+        response = self.client.post(path, post_data, follow=True)
+        self.assertEqual(response.status_code, 200)
```
nemesisdesign: and here

openwisp_users/tests/test_admin.py
```diff
@@ -493,6 +493,44 @@ def test_organization_uuid_field(self):
...
+
+    def test_action_confirmation_page(self):
+        user = User.objects.create(is_active=True)
```
nemesisdesign: same here

openwisp_users/tests/test_admin.py
```diff
@@ -493,6 +493,44 @@ def test_organization_uuid_field(self):
         html = '<input type="text" name="name" value="default"'
         self.assertNotContains(response, html)
 
+    def test__action_active(self):
+        user = User.objects.create(is_active=False)
+        path = reverse('admin:openwisp_users_user_changelist')
+        self.client.force_login(self._get_admin())
+        post_data = {
```
nemesisdesign: indentation looks weird, please make sure it's like:
```python
post_data = {
    'key': 'value'
}
```
same applies for the other test methods

openwisp_users/tests/test_admin.py
nepython: @nemesisdesign, thanks for suggesting this, I have implemented it. I just realized even when the response status is 200, it might be that the action was not performed which as you said made the test useless.

openwisp_users/admin.py
nepython: Can you please check the docstring I have added, I am not good at documenting my changes but tried to make it as simple as possible.

nepython:
@nemesisdesign, can you please review this too :grinning:

nepython:
> Great work @nepython. Now the goal should be that in your next patches you can get to this point with a lot less interaction and review on our side. That's why you should keep training.

Sure, @nemesisdesign will try to achieve that
:)

PR #105, opened 2020-04-17
[api] Add API endpoint to authenticate and get a REST token #104

### LINKED ISSUE #34: [admin] Only organizationuser objects which have admin=True should be able to manage their organization in the admin

### LINKED ISSUE #104: [api] Add API endpoint to authenticate and get a REST token
See https://www.django-rest-framework.org/api-guide/authentication/#tokenauthentication.

This is an optional feature which should be disabled by default, the setting to use shall be `OPENWISP_USERS_AUTH_API`.

Add django-rest-framework in the `extra_requires['rest']` of setup.py. 

[Add an API endpoint](https://www.django-rest-framework.org/api-guide/authentication/#by-exposing-an-api-endpoint) which allows users to authenticate using `ObtainAuthTokenView`, with the difference that we will use the  "Bearer" keyword and we will throttle the requests to avoid that a single IP address can attempt to log in too many times (this should be configurable).

Let's not forget to mention this feature in the README.

PabloCastellano:
* API is disabled by default and can be enabled by setting
  OPENWISP_USERS_AUTH_API = True
* API authentication rate limit can be configured using
  OPENWISP_USERS_AUTH_THROTTLE_RATE
* Require Redis from now on, since cache is needed for rate limiting

Closes #104

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -76,6 +76,7 @@ Setup (integrate in an existing django project)
         'django.contrib.sessions',
         'django.contrib.messages',
         'django.contrib.staticfiles',
+        'openwisp_utils.admin_theme',
         'django.contrib.admin',
         'django.contrib.sites',
         'django_extensions',
@@ -83,6 +84,8 @@ Setup (integrate in an existing django project)
         'allauth.account',
         'allauth.socialaccount',
         'openwisp_users',
+        'rest_framework',
+        'rest_framework.authtoken',
     ]
 
 also add ``AUTH_USER_MODEL`` and ``SITE_ID`` to your ``settings.py``::
@@ -101,6 +104,7 @@ also add ``AUTH_USER_MODEL`` and ``SITE_ID`` to your ``settings.py``::
     urlpatterns = [
         url(r'^admin/', include(admin.site.urls)),
         url(r'^accounts/', include('allauth.urls')),
+        url(r'^api/', include('openwisp_users.api.urls')),
     ]
 
     urlpatterns += staticfiles_urlpatterns()
@@ -130,6 +134,12 @@ Install test requirements:
 .. code-block:: shell
 
     pip install -r requirements-test.txt
+
+Start Redis
+
+.. code-block:: shell
+
+    docker-compose up -d
 
 Create database:
 
@@ -193,6 +203,34 @@ is enabled or not.
 
 It is disabled by default because `OpenWISP <http://openwisp.org>`_ does not use
 this feature of `django-organizations <https://github.com/bennylope/django-organizations>`_ yet.
+
+``OPENWISP_USERS_AUTH_API``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
++--------------+--------------+
+| **type**:    | ``boolean``  |
++--------------+--------------+
+| **default**: | ``False``    |
++--------------+--------------+
+
+Indicates whether the API is enabled or not.
+
+``OPENWISP_USERS_AUTH_THROTTLE_RATE``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
++--------------+--------------+
+| **type**:    | ``str``      |
++--------------+--------------+
+| **default**: | ``100/day``  |
++--------------+--------------+
+
+Indicates the rate throttling for the API authentication endpoint.
+
+Please note that the current rate throttler is very basic and will
+also count valid requests for rate limiting. For more information,
+check Django-rest-framework
+`throttling guide <https://www.django-rest-framework.org/api-guide/throttling/>`_.
+
 
 Multitenancy mixins
 -------------------
--- a/README.rst
+++ b/README.rst
@@ -104,7 +104,7 @@ also add ``AUTH_USER_MODEL`` and ``SITE_ID`` to your ``settings.py``::
     urlpatterns = [
         url(r'^admin/', include(admin.site.urls)),
         url(r'^accounts/', include('allauth.urls')),
-        url(r'^api/', include('openwisp_users.api.urls')),
+        url(r'^api/v1/', include('openwisp_users.api.urls')),
     ]
 
     urlpatterns += staticfiles_urlpatterns()
--- a/openwisp_users/api/tests/test_authentication.py
+++ b/openwisp_users/api/tests/test_authentication.py
@@ -1,0 +1,30 @@
+from django.core.cache import cache
+from django.test import RequestFactory, TestCase
+from openwisp_users.api.authentication import BearerAuthentication
+from openwisp_users.tests.utils import TestMultitenantAdminMixin
+from rest_framework.decorators import api_view, authentication_classes
+from rest_framework.response import Response
+
+from . import AuthenticationMixin
+
+
+class AuthenticationTests(TestMultitenantAdminMixin, AuthenticationMixin, TestCase):
+    def setUp(self):
+        cache.clear()
+        self.factory = RequestFactory()
+        self.operator = self._create_operator()
+
+    def test_bearer_authentication(self):
+        @api_view(['GET'])
+        @authentication_classes([BearerAuthentication])
+        def my_view(request):
+            return Response({})
+
+        request = self.factory.get('/')
+        response = my_view(request)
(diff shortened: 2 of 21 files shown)


PabloCastellano:
@nemesisdesign I have finally removed this piece because it is not needed for openwisp-users. 

```
        'DEFAULT_AUTHENTICATION_CLASSES': [                                                                                                                                                                                             
            'openwisp_users.api.authentication.BearerAuthentication',                                                                                                                                                                   
        ],                                                                                                                                                                                                                              
```

We will add it in openwisp-firmware-upgrader.

openwisp_users/apps.py
```diff
@@ -1,16 +1,24 @@
...
 
 
 class OpenwispUsersConfig(AppConfig):
     name = 'openwisp_users'
     app_label = 'openwisp_users'
     verbose_name = _('Users and Organizations')
 
+    rest_framework_defaults = {
+        'DEFAULT_THROTTLE_RATES': {
+            'anon': app_settings.USERS_AUTH_THROTTLE_RATE,
```
nemesisdesign, line 14: I'm not convinced of this.

Can't we just set this rate directly to our view and leave this setting on its own?

That way we can lower the default throttling rate for this endpoint, I think 20 requests a day should be more than enough. But we can't set such a low limit for all the endpoints.

PabloCastellano: Sounds good, I also had the same feeling.

PabloCastellano:
Made the requested changes. I think it is a good idea to keep BearerAuthentication in openwisp-users, it looks to me this way it is easier to reuse in other projects. WDYT?

openwisp_users/apps.py
```diff
@@ -1,16 +1,22 @@
...
 
 
 class OpenwispUsersConfig(AppConfig):
     name = 'openwisp_users'
     app_label = 'openwisp_users'
     verbose_name = _('Users and Organizations')
 
+    rest_framework_defaults = {
+        'DEFAULT_PERMISSION_CLASSES': ['rest_framework.permissions.IsAuthenticated']
+    }
```
nemesisdesign, line 14: yes I think we can set the bearer authentication as default here, is that what you meant? Although I'm a bit worried it may affect the existing APIs of some modules.

PabloCastellano: BearerAuthentication class is not used so far in openwisp-users because it makes no sense to require the bearer token in order to get it from the endpoint.

BearerAuthentication class is used in openwisp-firmware-upgrader and maybe other projects. Where do we want to place this class that all projects requiring api authentication can just import it and so avoid duplicating it?

nemesisdesign: I think openwisp-users is the right place because here is where we will deal with everything related to the management of users, organizations, multi-tenancy, permissions and I'd say API authentication is related. What do you think?

PabloCastellano: Yes, I also think this is the correct place. Do you think we need a test for it if it's not used in this module?

nemesisdesign: Yes we do, so that other modules don't have to write a test for it and they can assume it works (like we do with other libraries).

You can add a dummy API endpoint in the test project: https://github.com/openwisp/openwisp-users/tree/master/tests/testapp

PabloCastellano:
Done. I have added a test for BearerAuthentication and another one for AuthRateThrottle

tests/urls.py
```diff
@@ -5,6 +5,7 @@
 urlpatterns = [
     url(r'^admin/', admin.site.urls),
     url(r'^accounts/', include('openwisp_users.accounts.urls')),
+    url(r'^api/', include('openwisp_users.api.urls')),
```
nemesisdesign, line 8: let's add v1: `url(r'^api/v1/', include('openwisp_users.api.urls')),`

openwisp_users/api/urls.py
```diff
@@ -0,0 +1,11 @@
...
+from openwisp_users import settings as app_settings
+
+from . import views
+
+urlpatterns = []
+
+if app_settings.USERS_AUTH_API:
+    urlpatterns += [
+        url(r'^users/token/', views.obtain_auth_token, name='api-token-auth')
+    ]
```
nemesisdesign, line 11: for consistency with the other modules, let's use underscores instead of dashes in the `name` argument.

Similar endpoint in django-freeradius: https://github.com/openwisp/django-freeradius/blob/master/django_freeradius/api/urls.py#L15-L20

Let's call it `user_auth_token` as well, we'll separate the two using URLs namespaces (the URLs in the radius module are already namespaced)

openwisp_users/api/urls.py
```diff
@@ -0,0 +1,11 @@
+from django.conf.urls import url
+from openwisp_users import settings as app_settings
+
+from . import views
+
+urlpatterns = []
+
```
nemesisdesign, line 7: I gave it some more thought and concluded it will be better if we namespace the URLs (as we do in other modules), can we add `app_name = 'users'` to divide these endpoints in this namespace please?

openwisp_users/apps.py
nemesisdesign: I have one last doubt regarding this default here, wouldn't this make any DRF view require authentication by default?
If yes, this will affect other modules which already have APIs.

openwisp_users/api/tests/test_authentication.py
```diff
@@ -0,0 +1,30 @@
...
+        @authentication_classes([BearerAuthentication])
+        def my_view(request):
+            return Response({})
+
+        request = self.factory.get('/')
+        response = my_view(request)
+        self.assertEqual(response.status_code, 401)
+
+        token = self._obtain_auth_token()
+        request = self.factory.get('/', HTTP_AUTHORIZATION='Bearer ' + token)
```
nemesisdesign, line 28: try this: `f'Bearer {token}'`

openwisp_users/api/tests/test_authentication.py
```diff
@@ -0,0 +1,30 @@
...
+from django.test import RequestFactory, TestCase
+from openwisp_users.api.authentication import BearerAuthentication
+from openwisp_users.tests.utils import TestMultitenantAdminMixin
+from rest_framework.decorators import api_view, authentication_classes
+from rest_framework.response import Response
+
+from . import AuthenticationMixin
+
+
+class AuthenticationTests(TestMultitenantAdminMixin, AuthenticationMixin, TestCase):
```
nemesisdesign, line 11: for consistency with the other modules, we should move these tests to `openwisp_users.tests.test_api`.

I'd merge the 3 classes into one since there's not many lines in each class so one file and one class for all these should be well manageable

README.rst
```diff
@@ -101,6 +104,7 @@ also add ``AUTH_USER_MODEL`` and ``SITE_ID`` to your ``settings.py``::
     urlpatterns = [
         url(r'^admin/', include(admin.site.urls)),
         url(r'^accounts/', include('allauth.urls')),
+        url(r'^api/', include('openwisp_users.api.urls')),
```
nemesisdesign, line 107: v1 here as well

openwisp_users/apps.py
PabloCastellano: You are correct. I was not aware that other modules have already an API, so probably the best here is to add the class individually to the views that need it. Maybe we can do it in a Mixin?

nemesisdesign: @PabloCastellano I think we can remove this from here and forget it for now. Consider this the first iteration in this module, we'll surely come back here in the near future with clearer idea of how to improve it.

At some point we'll have to add endpoints to manage the models of this module (User, Organization, Group, etc) as well, but we'll move with this gradually.

PR #112, opened 2020-04-30
[admin] Superuser can't delete a regular user #110

### LINKED ISSUE #110: Superuser can't delete a regular user
When trying to delete a regular user as admin (with superuser status) I get the following message:

grafik

But there's no user with more privileges.

This is openwisp-users 0.2.0.

NoumbissiValere:
It is but normal that a superuser should able to
delete any object including organizationuser.
Thus, I fixed the issue by updating the
has_delete_permision of organizationuser to
return True for superusers

Fixes #110

### DIFF
--- a/openwisp_users/admin.py
+++ b/openwisp_users/admin.py
@@ -362,7 +362,9 @@ class OrganizationUserAdmin(
         operators should not delete organization users of organizations
         where they are not admins
         """
-        if obj and not request.user.is_superuser:
+        if request.user.is_superuser:
+            return True
+        if obj:
             operator_org = OrganizationUser.objects.get(
                 organization=obj.organization, user=request.user
             )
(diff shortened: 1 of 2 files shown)


NoumbissiValere:
@okraits I think this fixes the issue

openwisp_users/admin.py
```diff
@@ -370,6 +370,8 @@ def has_delete_permission(self, request, obj=None):
                 return True
             else:
                 return False
+        if request.user.is_superuser:
+            return True
```
PabloCastellano, line 374: IMO the logic is clearer if we do:

```
if request.user.is_superuser:
    return True
if obj:
    ...
```

NoumbissiValere: That's true @PabloCastellano thanks for the review

PR #121, opened 2020-05-21
[helpers] Organization membership related methods #107

### LINKED ISSUE #107: [api] permission class for organization membership and related methods
Here's some things we'll surely need. Related to #104.

### DRF Permission Class to check organization membership

A DRF [permission class](https://www.django-rest-framework.org/api-guide/permissions/#djangomodelpermissions) that does the same type of check we did in the private storage of openwisp-firmware-upgrader: https://github.com/openwisp/openwisp-firmware-upgrader/blob/6740240b9b7121fc59e8b9538ccda6e306767c5a/openwisp_firmware_upgrader/private_storage/views.py#L17-L18.

But it will have to use the methods defined below.

### Method to get organizations the user is member of + caching

A method in the `User` model which will:

- produce a data structure representing the organizations the user is member of, eg:

```python
{
    '205e999b-1394-48c0-a966-f517fa5c606a': {
        'name': 'default',
        'is_admin': True
    },
    'bb9fd683-085f-4824-b24c-4093cdb79734': {
        'name': 'staging',
        'is_admin': False
    }
}
```
 - cache the data structure value (no need to serialize it because the cache backend will do this on its own, we just have to ensure it's a dict and not an object so that it will work also when using JSON serialization), with a default expiry of something like 2 days
- return the value from the cache if present, otherwise generate it

Additional steps:

- Implement cache invalidation: whenever an `OrganizationUser` object is saved or deleted, the cache for that user is deleted. We shall use `post_save` and `post_delete` signals which are more reliable than `save` and `delete` method on the model (because model methods can be skipped when doing certain django ORM operations)

### Methods to check membership

- A method on the `User` model, which can easily check if the user can manage part of the objects in an organization, eg: `is_manager(organization)`, the check shall use the cached data structure defined above, eg:

```python
org_pk = str(organization.pk)
return (
    org_pk in self.organizations_data and 
    self.organizations_data[org_pk]['is_admin'] is True
)
```

:point_up_2: must ensure `is_admin is True`.

- A method on the `User` model, which can easily check if the user is member of an organization, eg: `is_member(organization)`, eg:

```python
org_pk = str(organization.pk)
return org_pk in self.organizations_data
```

:point_up_2: will check if the membership regardless of the value in `is_admin`.

- raise a `DeprecationWarning` exception in this older (and uglier) method, which will force us to [update the other modules](https://github.com/search?q=org%3Aopenwisp+organizations_pk&type=Code): https://github.com/openwisp/openwisp-users/blob/d95cb7a1e30ef69b948216c54931ddf7a4a215fc/openwisp_users/models.py#L57-L67 

- Document these methods in the README

- Ensure there's at least a test for each method

- Once this is ready, create an issue in openwisp-firmware-upgrader to use this new method: https://github.com/openwisp/openwisp-firmware-upgrader/blob/6740240b9b7121fc59e8b9538ccda6e306767c5a/openwisp_firmware_upgrader/private_storage/views.py#L17-L18

This functionality has to be released in a new point release, 0.3.0, in order to not break modules using `openwisp-users ~= 0.2` (the DeprecationWarning exception will probably break the other modules, but that's on purpose so we'll be forced to upgrade that).

PabloCastellano:
The User model provides the following methods to check whether the user
is a member or an administrator of an organization:

* `def is_manager(self, organization)`
* `def is_member(self, organization)`

Closes #107


openwisp_users/models.py
```diff
@@ -64,6 +68,11 @@ def organizations_pk(self):
         """
         returns primary keys of organizations the user is associated to
         """
+        warnings.warn(
+            "User.organizations_pk is deprecated in favor of User.organizations_dict"
+            " and will be removed in a future version",
+            DeprecationWarning,
+        )
```
nemesisdesign, line 75: could you try:

```
import logging
logging.getLogger(__name__)
logging.warn(msg)
```

PR #123, opened 2020-05-31
[improvement] Add swappable models #122

### LINKED ISSUE #122: [swappable] Add swappable models
Very often, organizations using OpenWISP need to customize the user model to add a custom field that they need or want.

For example, social security number, every country has a similar field with different length and validation rules.
Some organizations want to know the gender of the user or other details.

Some companies may be serving multiple customers and they may want to add more information to the "Organization" model, maybe they are using a CRM system and they want openwisp to link it, but to do so they need to store the ID of the customer in the CRM system in OpenWISP.

For all these reasons, it would be great to do the work we've been doing to add swappable models and extensible classes and the related documentation in this module as well.

We should also make sure that when the user model is swapped, if an additional required field is added, it will be included in the user creation page, which is handled by the [UserCreationForm](https://github.com/openwisp/openwisp-users/blob/master/openwisp_users/admin.py#L114-L127) (we'll need a test for this, likely in the SAMPLE_APP).

atb00ker:
Closes #122


openwisp_users/models.py
```diff
@@ -1,133 +1,41 @@
...
-    url = models.URLField(_('URL'), blank=True)
+        abstract = False
 
-    def __str__(self):
-        value = self.name
-        if not self.is_active:
-            value = '{0} ({1})'.format(value, _('disabled'))
-        return value
 
+class Organization(BaseOrganization, AbstractOrganization):
```
atb00ker, line 23: I had to use `class Organization(BaseOrganization, AbstractOrganization)` here because if I try to do something like:

```
from organizations.abstract import AbstractOrganization

class BaseOrganization(AbstractOrganization):
   # Module changes    
   class Meta:
         abstract = True

class Organization(BaseOrganization):
   class Meta:
         abstract = False
```

I get error stating `BaseOrganization` should not be abstract. 
I tried to find the reason in the docs of `django-organization` and I *think* if we want to make changes in Organzation's Abstract class, we need to use `[OrganizationBase](https://django-organizations.readthedocs.io/en/latest/cookbook.html?highlight=Abstract#difference-between-abstract-and-base-models)` which requires us to copy a lot of code from django-organization.
I tried to look into django-organization's code as well, but I can't find the line that's causing it to think `BaseOrganization` is the concrete class, hence this solution.

If you have any information about this and can give me pointers of where I should look to find the cause of the error, that will be very helpful.

tests/openwisp2/sample_users/models.py
```diff
@@ -0,0 +1,38 @@
...
+    pass
+
+
+class OrganizationOwner(BaseOrganizationOwner, AbstractOrganizationOwner):
+    pass
+
+
+# Boilerplate to add Permissions Groups
+class Group(BaseGroup, AbstractGroup):
+    pass
```
atb00ker, line 38: More boilerplate for our user! :-(
The problem here has been that there is no `AbstractGroup` class, and we are proxying a concrete class and it needs to be put in order to move `GroupAdmin` under the same app label as the other models.

atb00ker:
Tentatively applied these changes to openwisp-radius and tested, found some improvements and I think now it's ready for review.

README.rst
```diff
@@ -310,6 +310,326 @@ Multitenancy mixins
...
+You have to replicate and adapt that code in your project.
+
+
+For more information regarding the concept of ``AppConfig`` please refer to
+the `"Applications" section in the django documentation <https://docs.djangoproject.com/en/dev/ref/applications/>`_.
+
+7. Create your custom models
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+For the purpose of showing an example, we added a simple "fiscal_number" field in User model to the
```
nemesisdesign, line 443: This is great!

I am just not convinced of "fiscal number", is it a term you have heard before? In Italy it's called "Codice fiscale" but I don't think other nationalities will get it.

What about "Social Security Number"? Spain has "Seguridad Social" which is very close so I suspect this term may be understood by a broader audience.

We could then write something like:

```
This is the classic example of field that varies a lot from country to country, every country has a different name and different validation rules, so that's why these are not included by default in OpenWISP and is left up to the user to implement them.
```

Some countries even have libraries or services to validate these IDs, so we may point out that custom validation and verification systems can be implemented as well.

In the example images, please use a [sample SSN like the ones shown in google images](https://www.google.com/search?q=social+security+number+example&client=ubuntu&hs=Une&sxsrf=ALeKk00-fPXIb5gpuxbv4CwiuZ2NmmQn_w:1591380396346&source=lnms&tbm=isch&sa=X&ved=2ahUKEwjUuOf5oevpAhU2IrkGHSLjD_UQ_AUoAXoECA4QAw&biw=1920&bih=863#imgrc=RhJ7I8-HJbnWDM), to make the example more realistic.

openwisp_users/utils.py
```diff
@@ -0,0 +1,44 @@
...
+            fieldsets[: field[0]] + [field[1]] + fieldsets[field[0] :]
+        )
+
+
+def usermodel_list_and_search(model, additional_fields):
+    # Change form fieldsets
+    for field in additional_fields:
+        displays = model.list_display[:]
+        model.list_display = displays[: field[0]] + [field[1]] + displays[field[0] :]
+        model.search_fields += (field[1],)
```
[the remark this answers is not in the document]
nemesisdesign: maybe it would help if this file was in base/admin.py?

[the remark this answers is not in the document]
atb00ker: I've documented them in the README : https://github.com/openwisp/openwisp-users/pull/123/files#diff-88b99bb28683bd5b7e3a204826ead112R509-R527

I'll link that documentation here for the developers as well! :smile:

README.rst
atb00ker: Okay, I'll do that.
How about a field name like "Employee Number" or "College Enrollment Number"; conveys the message that a custom field is added and I think everyone will get that.

atb00ker: > In the example images, please use a sample SSN like the ones shown in google images, to make the example more realistic.

I didn't get that, do you mean, I should add an image of this in the README?

openwisp_users/utils.py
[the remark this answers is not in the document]
atb00ker: > maybe it would help if this file was in base/admin.py?

It's not really used in the `base/admin.py` and is optional to be used when extending, hence I added it here.

Should I have this and 

```

if 'reversion' in settings.INSTALLED_APPS:  # pragma: no cover
    from reversion.admin import VersionAdmin as BaseModelAdmin
else:
    from django.contrib.admin import ModelAdmin as BaseModelAdmin


class BaseAdmin(BaseModelAdmin):
    history_latest_first = True
```

with it or just the new functions?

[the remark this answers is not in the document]
nemesisdesign: Ok cool, let's not move it, please just a brief comment regarding what this file is for, like you did in the README but shorted or point to the README.

README.rst
nemesisdesign: > Okay, I'll do that.
> 
> > In the example images, please use a sample SSN like the ones shown in google images, to make the example more realistic.
> 
> I didn't get that, should I add a link to an image of this in the README?

I meant to use a realistic SSN and take the screenshots using the realistic SSN (instead of 1234)

> How about a field name like "Employee Number" or "College Enrollment Number"; conveys the message that a custom field is added and I think everyone will get that.

I like SSN because it's one of those things that people interested in OpenWISP ask.

openwisp_users/utils.py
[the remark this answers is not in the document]
atb00ker: Already done, there is a doc string in the functions, which points to the README like:
```
    """
    Read:
    https://github.com/openwisp/openwisp-users/blob/master/README.rst#usermodel_add_form
    """
```

PR #133, opened 2020-07-03
[admin] Update UserCreationForm #129

### LINKED ISSUE #103: [admin] Specify group during user creation
During creation, when a user is flagged as staff but not flagged as superuser, we should force the selection of a group (operator or administrator).

### LINKED ISSUE #129: [admin] Allow specifying more fields during user creation
To be done after or in conjunction with #103 (can be done in the same PR if preferred).

It would be nice to allow specifying these fields in the add form:

- first name
- last name
- phone number

NoumbissiValere:
Fixed the problem by updating the User
Creation form and also added a clean
method to user change form to validate
that a group must be selected for a staff
which Fixes #103

Fixes #129

### DIFF
--- a/openwisp_users/admin.py
+++ b/openwisp_users/admin.py
@@ -15,6 +15,7 @@ from django.contrib.auth.admin import GroupAdmin as BaseGroupAdmin
 from django.contrib.auth.admin import UserAdmin as BaseUserAdmin
 from django.contrib.auth.forms import UserChangeForm as BaseUserChangeForm
 from django.contrib.auth.forms import UserCreationForm as BaseUserCreationForm
+from django.core.exceptions import ValidationError
 from django.forms.models import BaseInlineFormSet
 from django.http import HttpResponseRedirect
 from django.template.response import TemplateResponse
@@ -27,6 +28,7 @@ from organizations.base_admin import (
     BaseOrganizationOwnerAdmin,
     BaseOrganizationUserAdmin,
 )
+from phonenumber_field.formfields import PhoneNumberField
 from swapper import load_model
 
 from . import settings as app_settings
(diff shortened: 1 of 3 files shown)


openwisp_users/utils.py
```diff
@@ -57,3 +59,12 @@ def usermodel_list_and_search(model, additional_fields):
         displays = model.list_display[:]
         model.list_display = displays[: field[0]] + [field[1]] + displays[field[0] :]
         model.search_fields += (field[1],)
+
+
+def validate_user_groups(data):
```
nemesisdesign, line 64: please move this to `admin.py`, we can use a mixin.

`EmailRequiredMixin` is currently being inherited by both forms, so we could do change that to:

```python
class UserFormMixin(forms.ModelForm):
    email = forms.EmailField(label=_('Email'), max_length=254, required=True)

    def validate_user_groups(data):
        is_staff = data.get('is_staff')
        groups = data.get('groups')
        if is_staff and not groups:
            raise ValidationError(
                {'groups': _('A staff must belong to a group, Please select one')}
            )

     def clean(self):
        cleaned_data = super().clean()
        validate_user_groups(cleaned_data)
        return cleaned_data
```

openwisp_users/admin.py
```diff
@@ -122,16 +124,51 @@ def has_add_permission(self, request, obj=None):
...
+                    'classes': ('wide',),
+                    'fields': ['is_active', 'is_staff', 'is_superuser', 'groups'],
+                },
+            ),
         )
 
+    def clean(self):
+        cleaned_data = super().clean()
+        self.validate_user_groups(cleaned_data)
+        return cleaned_data
```
nemesisdesign, line 170: there's repetition of this in both forms, you should not do that.
See again my previous review, I left a suggestion that will solve this:
https://github.com/openwisp/openwisp-users/pull/133#pullrequestreview-442572930

openwisp_users/tests/test_admin.py
```diff
@@ -507,9 +542,85 @@ def test_admin_add_user_org_required(self):
...
+                f'{self.app_label}_organizationuser-MAX_NUM_FORMS': 1,
+            }
+        )
+        res = self.client.post(
+            reverse(f'admin:{self.app_label}_user_add'), params, follow=True
+        )
+        self.assertEqual(res.status_code, 200)
+        self.assertContains(
+            res, 'A staff user must belong to a group, please select one.'
+        )
```
nemesisdesign, line 580: also here query the DB to double check user has been created

openwisp_users/tests/test_admin.py
```diff
@@ -507,9 +542,85 @@ def test_admin_add_user_org_required(self):
...
+                f'{self.app_label}_organizationuser-MAX_NUM_FORMS': 1,
+            }
+        )
+        res = self.client.post(
+            reverse(f'admin:{self.app_label}_user_add'), params, follow=True
+        )
+        self.assertEqual(res.status_code, 200)
+        self.assertNotContains(res, 'error')
+        user = User.objects.filter(username='testadd')
+        self.assertEqual(user.count(), 1)
```
nemesisdesign, line 611: here you did it, why in the other tests you didn't?

openwisp_users/tests/test_admin.py
```diff
@@ -507,9 +542,85 @@ def test_admin_add_user_org_required(self):
...
+        )
+        res = self.client.post(
+            reverse(f'admin:{self.app_label}_user_add'), params, follow=True
+        )
+        self.assertEqual(res.status_code, 200)
+        self.assertContains(
+            res, 'A staff user must belong to a group, please select one.'
+        )
+
+    def test_add_user_with_group(self):
```
nemesisdesign, line 582: instead of testing this, can you please test a non staff and non superuser?
We expect the user to be created normally in this case

PR #139, opened 2020-07-08
[admin] Make organization owner read-only #132

### LINKED ISSUE #132: [admin] Make organization owner read-only to other organization admins
If a user is organization owner and another organization administrator is viewing its change user page, make everything read-only.

We can override `get_readonly_fields` for this.

NoumbissiValere:
Closed issued by updating the change_permissions
of UserAdmin, OrganizationUserInline and
EmailAddressInline to meet the required needs.

Fixes #132

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,6 +8,7 @@ WIP
 
 - `organization owners <https://github.com/openwisp/openwisp-users#organization-owners>`_:
   [admin] Added support for organization owners
+- [admin] Made organization owners read-only to non-superusers
 
 Version 0.2.2 [2020-05-04]
 --------------------------
--- a/openwisp_users/admin.py
+++ b/openwisp_users/admin.py
@@ -58,6 +58,12 @@ class EmailAddressInline(admin.StackedInline):
         """
         return False
 
+    def has_change_permission(self, request, obj=None):
+        if user_not_allowed_to_change_owner(request.user, obj):
+            self.can_delete = False
+            return False
+        return super().has_change_permission(request, obj)
+
 
 class RequiredInlineFormSet(BaseInlineFormSet):
     """
@@ -124,6 +130,11 @@ class OrganizationUserInlineReadOnly(OrganizationUserInline):
 
     def has_add_permission(self, request, obj=None):
         return False
+
+    def has_change_permission(self, request, obj=None):
+        if user_not_allowed_to_change_owner(request.user, obj):
+            return False
+        return super().has_change_permission(request, obj)
 
 
 class UserFormMixin(forms.ModelForm):
@@ -203,6 +214,9 @@ class UserAdmin(MultitenantAdminMixin, BaseUserAdmin, BaseAdmin):
     inlines = [EmailAddressInline, OrganizationUserInline]
     save_on_top = True
     actions = ['delete_selected_overridden', 'make_inactive', 'make_active']
+
+    # To ensure extended apps use this template.
+    change_form_template = 'admin/openwisp_users/user/change_form.html'
 
     def require_confirmation(func):
         """
@@ -312,6 +326,8 @@ class UserAdmin(MultitenantAdminMixin, BaseUserAdmin, BaseAdmin):
         return fields
 
     def has_change_permission(self, request, obj=None):
+        if user_not_allowed_to_change_owner(request.user, obj):
+            return False
         # do not allow operators to edit details of superusers
         # returns 403 if trying to access the change form of a superuser
         if (
@@ -319,6 +335,11 @@ class UserAdmin(MultitenantAdminMixin, BaseUserAdmin, BaseAdmin):
         ):  # pragma: no cover
             return False
         return super().has_change_permission(request, obj)
+
+    def has_delete_permission(self, request, obj=None):
+        if user_not_allowed_to_change_owner(request.user, obj):
+            return False
+        return super().has_delete_permission(request, obj)
 
     def get_actions(self, request):
         actions = super().get_actions(request)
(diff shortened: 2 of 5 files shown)


openwisp_users/admin.py
```diff
@@ -58,6 +58,14 @@ def has_add_permission(self, *args, **kwargs):
         """
         return False
 
+    def has_change_permission(self, request, obj=None):
+        if obj and request.user.pk != obj.pk:
```
nemesisdesign, line 62: If the user is superuser, things can work as before without the need of doing any query

openwisp_users/utils.py
```diff
@@ -57,3 +62,9 @@ def usermodel_list_and_search(model, additional_fields):
         displays = model.list_display[:]
         model.list_display = displays[: field[0]] + [field[1]] + displays[field[0] :]
         model.search_fields += (field[1],)
+
+
+def get_owned_organization(user):
+    return OrganizationOwner.objects.filter(
+        organization_user__in=OrganizationUser.objects.filter(user=user)
+    )
```
nemesisdesign, line 70: we can find out if the user is owner of any org efficiently by looping over `user.organization_dicts`, it's explained in [organizations_dict](https://github.com/openwisp/openwisp-users#organizations_dict), the output of that method is cached.

I would add a new helper property (use [cached_property](https://docs.djangoproject.com/en/3.0/ref/utils/#django.utils.functional.cached_property)) on the user model for this check, eg: `user.is_owner_of_any_organization`, if this method returns `True` you can then make the change form read only.

NoumbissiValere:
Thanks @nemesisdesign I will work on this

CHANGES.rst
```diff
@@ -6,6 +6,8 @@ Version 0.3.0 [unreleased]
 
 WIP
 
+- [admin] Make organization owner read-only
```
nemesisdesign, line 9: use always the past tense

openwisp_users/tests/test_admin.py
```diff
@@ -511,6 +511,119 @@ def test_change_staff_with_group(self):
...
+        params.update(self._additional_params_add())
+        params.update(self._get_edit_form_inline_params(user2, org))
+        path = reverse(f'admin:{self.app_label}_user_change', args=[user2.pk])
+        r = self.client.post(path, params, follow=True)
+        self.assertEqual(r.status_code, 200)
+        self.assertNotContains(r, 'error')
+        user2.refresh_from_db()
+        self.assertEqual(user2.username, 'newuser1')
+
+    def test_staff_can_edit_org_owner(self):
```
nemesisdesign, line 544: If I'm not mistaken, this test ensures a user can't edit an org owner.

I think this should be named `test_staff_cannot_edit_org_owner` and should be the first test of this group because it's the most important one.

openwisp_users/tests/test_admin.py
```diff
@@ -511,6 +511,119 @@ def test_change_staff_with_group(self):
...
+        params = user.__dict__
+        params['username'] = 'newusername'
+        params['groups'] = str(group.pk)
+        params.pop('phone_number')
+        params.pop('_password')
+        params.pop('last_login')
+        params.update(self.add_user_inline_params)
+        params.update(self._additional_params_add())
+        params.update(self._get_edit_form_inline_params(user, org))
+        path = reverse(f'admin:{self.app_label}_user_change', args=[user.pk])
```
nemesisdesign, line 620: there is a lot of repetition in these tests which is making the test code time consuming to read, can you please eliminate the repetitive lines and keep only the essential bits in each test?

openwisp_users/tests/test_admin.py
```diff
@@ -511,6 +511,119 @@ def test_change_staff_with_group(self):
...
+        r = self.client.post(path, params, follow=True)
+        self.assertEqual(r.status_code, 200)
+        self.assertNotContains(r, 'error')
+        user.refresh_from_db()
+        self.assertEqual(user.username, 'newusername')
+
```
nemesisdesign, line 626: I think one important case is missing from the tests: a staff user should be able to edit their own account.
This works now, but someone may inadvertently break this in the future, so if we add a test to ensure this works we prevent this future bug from happening.

openwisp_users/tests/test_admin.py
```diff
@@ -511,6 +511,119 @@ def test_change_staff_with_group(self):
...
+        )
+        self._create_org_user(user=user1, organization=org, is_admin=True)
+        group = Group.objects.filter(name='Administrator')
+        user1.groups.set(group)
+        user2.groups.set(group)
+        self.client.force_login(user1)
+        path = reverse(f'admin:{self.app_label}_user_change', args=[user2.pk])
+        r = self.client.get(path)
+        self.assertEqual(r.status_code, 200)
+        self.assertContains(r, f'class="readonly">{user2.username}')
```
nemesisdesign, line 564: please also check for the presence of the warning we just added (`You do not have permission to edit or delete this user because they are owner of an organization`).

openwisp_users/utils.py
```diff
@@ -1,10 +1,14 @@
+import logging
+
 from django.conf import settings
 
 if 'reversion' in settings.INSTALLED_APPS:  # pragma: no cover
     from reversion.admin import VersionAdmin as BaseModelAdmin
 else:
     from django.contrib.admin import ModelAdmin as BaseModelAdmin
 
+logger = logging.getLogger(__name__)
```
nemesisdesign, line 10: why is this being added again?
I had to remove it, it's causing issues, see https://github.com/openwisp/openwisp-users/commit/2e98874636621aa1d5406e293d63112de87c8c20

openwisp_users/base/models.py
```diff
@@ -88,6 +88,13 @@ def is_owner(self, organization):
         org_dict = self.organizations_dict.get(str(organization.pk))
         return org_dict is not None and org_dict['is_owner']
 
+    @cached_property
+    def is_owner_of_any_organization(self):
+        for org_pk, value in self.organizations_dict.items():
```
nemesisdesign, line 93: Since `org_pk` is not used, let's change this to:
```python
for value in self.organizations_dict.values():
    # ...
```
which is also cleaner.

NoumbissiValere:
:laughing: I thought I dealt with this during testing?

PR #140, opened 2020-07-08
[feature] Add support for organization owners #137

### LINKED ISSUE #137: [feature] Add support for organization owners
- when user is added to an organization (a new OrganizationUser with the `is_admin` flag `True` is created), if there's no owner yet, we can automatically add the user as owner, we could do this with a `post_save` signal.
- we can add a data migration that loops over all the organizations which do not have an organization owner defined and will add the older user (must have `OrganizationUser.is_admin` set to `True`) that was added (based on date in which the user account was created)
- we should add a way to easily manage owners, a possible way would be to add an inline to `OrganizationAdmin`

NoumbissiValere:
Closed issue  by adding a handler for the post_save
signal of OrganizationUser which creates an
organization owner with the first organization manager.
a migration file was also written to create default
organization owners for existing OrganizationUser
without organization owners. Lastly, an
OrganizationOwnerInline was added to OrganizationAdmin.

Fixes #137


openwisp_users/apps.py
```diff
@@ -71,3 +82,29 @@ def update_organizations_dict(cls, instance, **kwargs):
...
+                        organization_user=instance, organization=instance.organization
+                    )
+                    owner.full_clean()
+                    owner.save()
+                except ValidationError as e:
+                    logger.exception(
+                        'Got exception {} while saving OrganizationOwner with '
+                        'organization_user {} and organization {}'.format(
+                            type(e), instance, instance.organization
+                        )
```
nemesisdesign, line 109: we can use use f strings instead of `format`, since we only support python >= 3.6 now

NoumbissiValere: Ok

openwisp_users/admin.py
```diff
@@ -74,6 +74,10 @@ def _construct_form(self, i, **kwargs):
         return form
 
 
+class OrganizationOwnerInline(admin.StackedInline):
+    model = OrganizationOwner
```
nemesisdesign, line 78: add `extra = 0`

nemesisdesign: also here we should manipulate the permissions:

- add/change/delete: only owners and superusers

NoumbissiValere: @nemesisdesign I don't think we should add this here because the default permissions allow only superusers to see the `Organization` model. all other users can't even see it. But i think #143 will solve and meet this need

nemesisdesign: yes you're right

openwisp_users/apps.py
```diff
@@ -71,3 +82,28 @@ def update_organizations_dict(cls, instance, **kwargs):
         cache.delete(cache_key)
         # forces caching
         user.organizations_dict
+
+    def create_organization_owner(cls, instance, **kwargs):
+        if not kwargs.get('created') or not instance.is_admin:
+            return
+        owner_model = load_model('openwisp_users', 'OrganizationOwner')
+        model = kwargs.get('sender')
```
nemesisdesign, line 90: why don't you call these two variables simply `OrganizationOwner` and `OrganizationUser`? Wouldn't this make the code more self explanatory?

openwisp_users/migrations/__init__.py
```diff
@@ -111,3 +111,26 @@ def update_admins_permissions(apps, schema_editor):
...
+def create_organization_owners(apps, schema_editor):
+    org_owner_model_name = swapper.get_model_name('openwisp_users', 'OrganizationOwner')
+    org_owner_label = swapper.split(org_owner_model_name)[0]
+    org_owner_model = apps.get_model(org_owner_label, 'OrganizationOwner')
+    org_user_model_name = swapper.get_model_name('openwisp_users', 'OrganizationUser')
+    org_user_label = swapper.split(org_user_model_name)[0]
+    org_user_model = apps.get_model(org_user_label, 'OrganizationUser')
+    org_model_name = swapper.get_model_name('openwisp_users', 'organization')
+    org_model_label = swapper.split(org_model_name)[0]
+    org_model = apps.get_model(org_model_label, 'organization')
```
nemesisdesign, line 125: why don't you make a helper method that retrieves the model? So we avoid repeating `get_model_name`, `split` and `get_model` so many times, then you can simply do

```python
OrganizationOwner = get_model('OrganizationOwner')
```
and so on

openwisp_users/migrations/__init__.py
```diff
@@ -111,3 +111,26 @@ def update_admins_permissions(apps, schema_editor):
...
+    for org in org_model.objects.all():
+        if not org_owner_model.objects.filter(organization=org).exists():
```
nemesisdesign, line 127: use if exists continue to avoid nesting too much which makes the code harder to read

openwisp_users/tests/test_admin.py
```diff
@@ -692,6 +694,53 @@ def test_organization_owner(self):
...
+        org = self._get_org()
+        user = self._get_user()
+        org_user = self._create_org_user(organization=org, user=user, is_admin=True)
+        org_owner_qs = OrganizationOwner.objects.all()
+        self.assertEqual(org_owner_qs.count(), 1)
+        org_owner = org_owner_qs.first()
+        self.assertEqual(org_owner.organization, org)
+        self.assertEqual(org_owner.organization_user, org_user)
+
+    def test_first_orguser_creates_no_org_owner(self):
```
nemesisdesign, line 707: the point is not that it's first, the point is that it's not admin

openwisp_users/tests/test_admin.py
```diff
@@ -692,6 +694,53 @@ def test_organization_owner(self):
...
+    def test_view_org_owner_org_admin(self):
+        self.client.force_login(self._get_admin())
+        r = self.client.get(reverse(f'admin:{self.app_label}_organization_add'))
+        self.assertEqual(r.status_code, 200)
+        self.assertContains(r, 'Organization owners')
+
+    @patch.object(
+        OrganizationOwner, 'full_clean', side_effect=ValidationError('validation error')
+    )
+    def test_invalid_org_owner(self, mocked):
```
nemesisdesign, line 735: patch the exception method of logger and assert it has been called, see how it's done here:
https://github.com/openwisp/openwisp-network-topology/pull/73/files#diff-13c0cc68f2519b129b31a58d282842d4R112-R115

openwisp_users/apps.py
```diff
@@ -71,3 +82,28 @@ def update_organizations_dict(cls, instance, **kwargs):
         cache.delete(cache_key)
         # forces caching
         user.organizations_dict
+
+    def create_organization_owner(cls, instance, **kwargs):
```
nemesisdesign, line 86: you can add a `created` argument and avoid having to use `kwargs.get('created')` (so you can just use `created`) which will make the code easier to read

openwisp_users/apps.py
```diff
@@ -71,3 +82,28 @@ def update_organizations_dict(cls, instance, **kwargs):
...
+        if not kwargs.get('created') or not instance.is_admin:
+            return
+        owner_model = load_model('openwisp_users', 'OrganizationOwner')
+        model = kwargs.get('sender')
+        org_users = model.objects.filter(organization=instance.organization)
+        is_first = org_users.first() == org_users.last() == instance
```
nemesisdesign, line 92: is this how you're checking that the `OrganizationUser` just created is the first user in the org?

What about simply checking if there's any other `OrganizationUser` in this org apart from the one just created?

```python
OrganizationUser.objects.filter(organization=instance.organization).exclude(pk=instance.pk).exists()
```

This query should answer that question.

README.rst
```diff
@@ -375,6 +375,15 @@ Usage exmaple:
     >>> user.organizations_dict.keys()
     ... dict_keys(['20135c30-d486-4d68-993f-322b8acb51c4'])
 
+Organization Owners
+-------------------
+
+an organization owner is a user who is designated as the owner
```
nemesisdesign, line 381: Missing initial capital letter

PR #130, opened 2020-06-25
[admin] Allow administrators to delete other users #128

NoumbissiValere:
Fixed the issue by adding a has_delete_permission to
UserAdmin which returns true when the necessary
conditions are met. I Also updated the Administrators'
default permissions by adding email permissions to it

Fixes #128

PR #147, opened 2020-07-19
[sample-app] Unnecessary additional migration removed

atb00ker:


tests/openwisp2/sample_users/migrations/0002_default_groups_and_permissions.py
```diff
@@ -22,4 +23,7 @@ class Migration(migrations.Migration):
         migrations.RunPython(
             create_default_groups, reverse_code=migrations.RunPython.noop
         ),
+        migrations.RunPython(
+            update_admins_permissions, reverse_code=migrations.RunPython.noop
+        ),
```
atb00ker, line 28: This migration is required in the sample_app as it's referred in documentation.
However, following the pattern from other repositories, the additional migration is deleted. (Alternative would be to add that migration in documentation.)

atb00ker:
@NoumbissiValere Follow up on https://github.com/openwisp/openwisp-users/pull/130#discussion_r447901785

PR #151, opened 2020-07-22
[multitenancy] Organization Membership validators #150

### LINKED ISSUE #150: [api] Add reusable mechanism to validate organization membership
In each module where we have an API, we are implementing the validation of the organization membership of a user differently and this is not good.

We need a general and reusable solution for this.

I think we can use `permission_classes` for this and create some classes like:

`isOrganizationMember`, `isOrganizationManager`, `isOrganizationOwner`.

We have to allow specifying the name of the organization field, eg:

```python
permission_classes = (isOrganizationMember,)
organization_field = 'organization'  # default
```

In case the organization field is on a parent element, we should be able to do the following:

```python
permission_classes = (isOrganizationMember,)
organization_field = 'device__organization'
```

atb00ker:
Closes #150

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -378,6 +378,58 @@ Usage exmaple:
     >>> user.organizations_dict.keys()
     ... dict_keys(['20135c30-d486-4d68-993f-322b8acb51c4'])
 
+Django REST Framework Permission Classes
+----------------------------------------
+
+The custom `Django REST Framework <https://www.django-rest-framework.org/>`_
+permission classes ``IsOrganizationMember``, ``IsOrganizationManager``
+and ``IsOrganizationOwner`` can be used in the API to ensure that the
+request user is in the same organization as requested object and is
+organization member, manager or owner respectively. Usage example:
+
+.. code-block:: python
+
+    from openwisp_users.api.permissions import IsOrganizationManager
+    from rest_framework import generics
+
+    class MyApiView(generics.APIView):
+        permission_classes = (IsOrganizationMember,)
+
+``organization_field``
+~~~~~~~~~~~~~~~~~~~~~~
+
++--------------+------------------+
+| **type**:    | ``string``       |
++--------------+------------------+
+| **default**: | ``organization`` |
++--------------+------------------+
+
+``organization_field`` can be used to define where to look to
+find the organization of the current object.
+In most cases this won't need to be changed, but it does need to
+be changed when the ``organization`` is defined only on a parent object.
+
+For example, in `openwisp-firmware-upgrader <https://github.com/openwisp/openwisp-firmware-upgrader>`_,
+``organization`` is defined on ``Category`` and ``Build`` has a relation
+to ``category``, so the organization of Build instances is inferred from
+the organization of the Category.
+
+Therefore, to implement the permission class correctly, we would have to do:
+
+.. code-block:: python
+
+    from openwisp_users.api.permissions import IsOrganizationManager
+    from rest_framework import generics
+
+    class MyApiView(generics.APIView):
+        permission_classes = (IsOrganizationMember,)
+        organization_field = 'category__organization'
+
+This will translate into accessing ``obj.category.organization``.
+Ensure the queryset of your views make use of
+`select_related <https://docs.djangoproject.com/en/3.0/ref/models/querysets/#select-related>`_
+in these cases to avoid generating too many queries.
+
 Organization Owners
 -------------------
 
--- a/openwisp_users/api/permissions.py
+++ b/openwisp_users/api/permissions.py
@@ -1,0 +1,68 @@
+from django.utils.translation import gettext_lazy as _
+from rest_framework.permissions import BasePermission
+from swapper import load_model
+
+Organization = load_model('openwisp_users', 'Organization')
+
+
+class BaseOrganizationPermission(BasePermission):
+    def has_object_permission(self, request, view, obj):
+        organization = self.get_object_organization(view, obj)
+        return self.validate_membership(request.user, organization)
+
+    def has_permission(self, request, view):
+        return request.user and request.user.is_authenticated
+
+    def get_object_organization(self, view, obj):
+        organization_field = getattr(view, 'organization_field', 'organization')
+        fields = organization_field.split('__')
+        accessed_object = obj
+        for field in fields:
+            accessed_object = getattr(accessed_object, field, None)
+            if not accessed_object:
+                raise AttributeError(
+                    _(
+                        'Organization not found, `organization_field` '
+                        'not implemented correctly.'
+                    )
+                )
+        return accessed_object
+
+    def validate_membership(self, user, org):
+        raise NotImplementedError(
+            _(
+                'View\'s permission_classes not implemented correctly.'
+                'Please use one of the child classes: IsOrganizationMember, '
+                'IsOrganizationManager or IsOrganizationOwner.'
+            )
+        )
+
+
+class IsOrganizationMember(BaseOrganizationPermission):
+    message = _(
+        'User is not a member of the organization to which the '
+        'requested resource belongs.'
+    )
+
+    def validate_membership(self, user, org):
+        return org and (user.is_superuser or user.is_member(org))
+
+
+class IsOrganizationManager(BaseOrganizationPermission):
+    message = _(
+        'User is not a manager of the organization to which the '
+        'requested resource belongs.'
+    )
+
+    def validate_membership(self, user, org):
+        return org and (user.is_superuser or user.is_manager(org))
+
+
+class IsOrganizationOwner(BaseOrganizationPermission):
+    message = _(
+        'User is not a owner of the organization to which the '
+        'requested resource belongs.'
+    )
+
+    def validate_membership(self, user, org):
+        return org and (user.is_superuser or user.is_owner(org))
--- a/openwisp_users/tests/test_api/__init__.py
+++ b/openwisp_users/tests/test_api/__init__.py
@@ -7,8 +7,8 @@ class AuthenticationMixin:
     def _obtain_auth_token(self, username='operator', password='tester'):
         params = {'username': username, 'password': password}
         url = reverse('users:user_auth_token')
-        r = self.client.post(url, params)
-        return r.data["token"]
+        response = self.client.post(url, params)
+        return response.data['token']
 
 
 class APITestCase(TestMultitenantAdminMixin, AuthenticationMixin, TestCase):
--- a/runtests.py
+++ b/runtests.py
@@ -15,10 +15,7 @@ if __name__ == '__main__':
     if not os.environ.get('SAMPLE_APP', False):
         args.insert(2, 'openwisp_users')
         args.insert(3, 'testapp.tests')
-        args.insert(4, 'testapp.test_multitenancy')
     else:
         args.insert(2, 'openwisp2')
         args.insert(3, 'testapp.tests')
-        args.insert(4, 'testapp.test_multitenancy')
-    # args.insert(5, '-v 3')
     execute_from_command_line(args)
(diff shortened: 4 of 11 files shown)


openwisp_users/base/models.py
```diff
@@ -78,13 +78,19 @@ def organizations_pk(self):
...
 
     def is_manager(self, organization):
+        if self.is_superuser:
+            return True
         org_dict = self.organizations_dict.get(str(organization.pk))
         return org_dict is not None and (org_dict['is_admin'] or org_dict['is_owner'])
 
     def is_owner(self, organization):
+        if self.is_superuser:
+            return True
```
atb00ker, line 93: superuser can be owner, manager, member at will, so always return `True`.

nemesisdesign: superuser can do anything and these checks do not apply to them, `if user.is_superuser` should always be checked beforehand using these methods

openwisp_users/multitenancy.py
```diff
@@ -134,3 +135,47 @@ class MultitenantRelatedOrgFilter(MultitenantOrgFilter):
     """
 
     multitenant_lookup = 'organization__in'
+
+
+class BaseOrganizationPermission(object):
```
nemesisdesign, line 140: should not extend object but a base permission class in DRF, please read https://www.django-rest-framework.org/api-guide/permissions/#custom-permissions, you then have to define `has_object_permission`.

You'll get the organization from the object passed to `has_object_permission`, but we have to take into account the case in which the organization field is not on the object itself but on a parent, we need to define a way to support this.

atb00ker:
Okay, it's ready.
But it doesn't replace the `validate_membership` function in `openwisp_radius` where no obj is accessed.
Like `PasswordChangeView` or when an object is created like `BatchView`.

README.rst
```diff
@@ -375,6 +375,32 @@ Usage exmaple:
...
+------------------------------------------
+
+The custom rest_framework permission classes ``IsOrganizationMember``,
+``IsOrganizationManager`` and ``IsOrganizationOwner`` can be used in the
+API to ensure that the request user is in the same organization as requested
+object and is organization member, manager or owner respectively.
+
+Usage exmaple:
+
+.. code-block:: python
```
nemesisdesign, line 388: Please check how it looks:

https://github.com/atb00ker/openwisp-users/tree/issues/150-org-member-validator#organization-membership-permission_classes

openwisp_users/multitenancy.py
```diff
@@ -134,3 +136,52 @@ class MultitenantRelatedOrgFilter(MultitenantOrgFilter):
     """
 
     multitenant_lookup = 'organization__in'
+
+
+class BaseOrganizationPermission(BasePermission):
+    def get_organization(self, view, obj):
+        organization_field = getattr(view, 'organization_field', 'organization')
+        return getattr(obj, organization_field, None)
```
nemesisdesign, line 144: what if the organization field is on a parent object?

atb00ker: https://github.com/atb00ker/openwisp-users/tree/issues/150-org-member-validator#organization_field

README.rst
```diff
@@ -375,6 +375,55 @@ Usage exmaple:
     >>> user.organizations_dict.keys()
     ... dict_keys(['20135c30-d486-4d68-993f-322b8acb51c4'])
 
+Organization membership permission_classes
+------------------------------------------
+
+The custom rest_framework permission classes ``IsOrganizationMember``,
```
nemesisdesign, line 381: rest_framework > Django REST Framework (link also the website)

README.rst
```diff
@@ -375,6 +375,55 @@ Usage exmaple:
...
+Usage exmaple:
+
+.. code-block:: python
+
+    from openwisp_users.multitenancy import IsOrganizationManager
+
+    class MyApiView:
+        permission_classes = (IsOrganizationMember,)
+
+        def get(self, request, *args, **kwargs):
```
nemesisdesign, line 395: if the view extends APIView or any other DRF view, this should not be needed, hence we can provide the example pointing to the standard usage.

README.rst
```diff
@@ -375,6 +375,55 @@ Usage exmaple:
...
+            # Perform all the actions required and
+            # get the requested object in variable ``obj``
+            # and call check_object_permissions which returns
+            # ``403`` if the user is not allowed to access
+            # the requested object.
+            self.check_object_permissions(request, obj)
+
+organization_field
```
nemesisdesign, line 403:
```
``organization_field``
~~~~~~~~~~~~~~~~~~~~~~
```

README.rst
```diff
@@ -375,6 +375,55 @@ Usage exmaple:
     >>> user.organizations_dict.keys()
     ... dict_keys(['20135c30-d486-4d68-993f-322b8acb51c4'])
 
+Organization membership permission_classes
+------------------------------------------
```
nemesisdesign, line 379: I would change this title to something like `Django REST Framework Permission Classes`

README.rst
```diff
@@ -375,6 +375,55 @@ Usage exmaple:
...
+
+    from openwisp_users.multitenancy import IsOrganizationManager
+
+    class MyApiView:
+        permission_classes = (IsOrganizationMember,)
+        organization_field = 'object_parent__organization'
+
+This will get organization from object as ``obj.object_parent__organization``
+and check if the users is has membership in that organization.
+
```
nemesisdesign, line 426: Try to go more straight to the point and to provide a real world example: 

```
``organization_field`` can be used to define where to look to find the organization of the current object.

In most cases this won't need to be changed, but it does need to be changed when the ``organization`` is 
defined only on a parent object.

For example, in openwisp-firmware-upgrader (link), ``organization`` is defined on ``Category`` and ``Build`` has a relation to ``category``, so the organization of Build instances is inferred from the organization of the Category.

Therefore, to implement the permission class correctly, we would have to do:

......
organization_field = 'category__organization'

This will translate into accessing ``obj.category.organization``.
Ensure the queryset of your views make use of select_related (link django docs) in these cases to avoid generating too many queries.
```

openwisp_users/multitenancy.py
```diff
@@ -134,3 +136,52 @@ class MultitenantRelatedOrgFilter(MultitenantOrgFilter):
     """
 
     multitenant_lookup = 'organization__in'
+
+
+class BaseOrganizationPermission(BasePermission):
+    def get_organization(self, view, obj):
```
nemesisdesign, line 142: for consistency, I would rename it to ``get_object_organization``

openwisp_users/multitenancy.py
```diff
@@ -134,3 +136,52 @@ class MultitenantRelatedOrgFilter(MultitenantOrgFilter):
...
+    def validate_membership(self, user, org):
+        raise NotImplementedError(
+            'Please use one of the child classes: IsOrganizationMember,'
+            'IsOrganizationManager or IsOrganizationOwner'
+        )
+
+    def has_permission(self, request, view):
+        return request.user and request.user.is_authenticated
+
+    def has_object_permission(self, request, view, obj):
```
nemesisdesign, line 155: this method should be defined first to improve readability, then has_permission, get_object_organization and validate_membership last.

openwisp_users/multitenancy.py
```diff
@@ -134,3 +136,52 @@ class MultitenantRelatedOrgFilter(MultitenantOrgFilter):
...
+        organization = self.get_organization(view, obj)
+        return self.validate_membership(request.user, organization)
+
+
+class IsOrganizationMember(BaseOrganizationPermission):
+    message = (
+        'User is not a member of the organization to which the '
```
nemesisdesign, line 162: make this and the other strings translatable please

atb00ker:
Example usage: https://github.com/openwisp/openwisp-radius/pull/137

README.rst
```diff
@@ -378,6 +378,55 @@ Usage exmaple:
...
++--------------+------------------+
+| **default**: | ``organization`` |
++--------------+------------------+
+
+``organization_field`` can be used to define where to look to
+find the organization of the current object.
+In most cases this won't need to be changed, but it does need to
+be changed when the ``organization`` is defined only on a parent object.
+
+For example, in openwisp-firmware-upgrader (link), ``organization``
```
nemesisdesign, line 411: (link) should be converted to a link to the repo

README.rst
```diff
@@ -378,6 +378,55 @@ Usage exmaple:
...
+permission classes ``IsOrganizationMember``, ``IsOrganizationManager``
+and ``IsOrganizationOwner`` can be used in the API to ensure that the
+request user is in the same organization as requested object and is
+organization member, manager or owner respectively. Usage example:
+
+.. code-block:: python
+
+    from openwisp_users.multitenancy import IsOrganizationManager
+
+    class MyApiView(`rest_framework.generics.APIView <https://www.django-rest-framework.org/api-guide/generic-views/#generic-views>`_):
```
nemesisdesign, line 394: this link here is odd

openwisp_users/multitenancy.py
```diff
@@ -134,3 +136,52 @@ class MultitenantRelatedOrgFilter(MultitenantOrgFilter):
...
+class BaseOrganizationPermission(BasePermission):
+    def has_object_permission(self, request, view, obj):
+        organization = self.get_object_organization(view, obj)
+        return self.validate_membership(request.user, organization)
+
+    def has_permission(self, request, view):
+        return request.user and request.user.is_authenticated
+
+    def get_object_organization(self, view, obj):
+        organization_field = getattr(view, 'organization_field', 'organization')
```
nemesisdesign, line 150: This requirement that is in the original issue is not satisfied:

> In case the organization field is on a parent element, we should be able to do the following:
> 
> ```python
> permission_classes = (isOrganizationMember,)
> organization_field = 'device__organization'
> ```

How can we specify to look on the parent object with the following implementation?

README.rst
atb00ker: Fixed.

README.rst
```diff
@@ -378,6 +378,55 @@ Usage exmaple:
...
+
+    from openwisp_users.multitenancy import IsOrganizationManager
+
+    class MyApiView:
+        permission_classes = (IsOrganizationMember,)
+        organization_field = 'category__organization'
+
+This will translate into accessing ``obj.category.organization``.
+Ensure the queryset of your views make use of `select_related <https://docs.djangoproject.com/en/3.0/ref/models/querysets/#select-related>`_
+in these cases to avoid generating too many queries.
```
atb00ker, line 428: Removing this line, as [the query, with or without select_related](https://explain.depesz.com/s/tbTt) remains the same.

openwisp_users/api/permissions.py
```diff
@@ -0,0 +1,65 @@
...
+            if not accessed_object:
+                self.message = _(
+                    'Organization not found, `organization_field` '
+                    'not implemented correctly.'
+                )
+                break
+        return accessed_object
+
+    def validate_membership(self, user, org):
+        self.message = _(
```
nemesisdesign, line 31: this should raise [NotImplementedError](https://docs.python.org/3/library/exceptions.html#NotImplementedError) with this message instead of trying to display it to the user

openwisp_users/api/permissions.py
```diff
@@ -0,0 +1,65 @@
...
+        return request.user and request.user.is_authenticated
+
+    def get_object_organization(self, view, obj):
+        organization_field = getattr(view, 'organization_field', 'organization')
+        fields = organization_field.split('__')
+        accessed_object = obj
+        for field in fields:
+            accessed_object = getattr(accessed_object, field, None)
+            if not accessed_object:
+                self.message = _(
```
nemesisdesign, line 23: if this happens, we should raise an exception, probably `AttributeError`, with the message used here.
If it's being used wrongly as in this case, it's better to let it fail and loud rather than work around it.

atb00ker: Ah, that's what I was doing before, but then I thought, letting user know. I've undone that.

openwisp_users/api/permissions.py
atb00ker: Ah, that's what I was doing before, but then I thought, letting user know. I've undone that.

PR #153, opened 2020-07-25
[model] cache user permissions #145

### LINKED ISSUE #145: [model] cache user permissions
create a new helper for the user model which caches user permissions.

NoumbissiValere:
Closed issue by adding a helper to
User model which caches all
permissions for a particular user.

Fixes #145


openwisp_users/base/models.py
```diff
@@ -129,6 +129,17 @@ def clean(self):
         if self.phone_number == '':
             self.phone_number = None
 
+    @cached_property
+    def get_permissions(self):
```
nemesisdesign, line 133: if it's a property, we don't need to name it `get_permissions`, we can name it just `permissions`.

The other problem is that `cached_property` does not cache using the cache storage, it caches only in memory, so the query will be executed again at subsequent requests.

We have to do something like `organizations_dict`: cache it in the cache storage (one key for each user) and invalidate the cache whenever the permissions or the groups of a user are changed.

Please study this PR and try to replicate what has been done there: https://github.com/openwisp/openwisp-users/pull/134/files

openwisp_users/base/models.py
```diff
@@ -129,6 +129,17 @@ def clean(self):
...
+        return self.get_all_permissions()
+
+    def refresh_from_db(self, *args, **kwargs):
+        super().refresh_from_db(*args, **kwargs)
+        try:
+            del self.__dict__['get_permissions']
+        except KeyError:
+            pass
```
nemesisdesign, line 141: why this? Looks like something that shouldn't be here.

openwisp_users/tests/test_models.py
```diff
@@ -181,3 +183,22 @@ def test_add_users_with_empty_phone_numbers(self):
...
+        user.groups.set(group)
+
+        self.assertEqual(user.get_all_permissions(), user.get_permissions)
+
+        user.groups.set(Group.objects.filter(name='Administrator'))
+        self.assertEqual(user.get_all_permissions(), user.get_permissions)
+
+        permission = Permission.objects.filter(codename='add_organization')
+        user.user_permissions.add(*permission)
+        self.assertEqual(user.get_all_permissions(), user.get_permissions)
```
nemesisdesign, line 204: please use assertNumQueries to ensure the query is executed only at the first call

NoumbissiValere:
ok @nemesisdesign thanks for the feedback. let me adjust them :smile:

README.rst
```diff
@@ -299,6 +299,12 @@ the token in the ``Authorization`` header:
     # send bearer token
     http GET localhost:8000/api/v1/firmware/build/ "Authorization: Bearer $TOKEN"
 
+User permissions
+---------------
```
nemesisdesign, line 303: the length of this must be equal to the length of the heading.
I would also move this section after the organization membership helpers and call it `User permissions helper` for consistency.

openwisp_users/base/models.py
```diff
@@ -129,6 +129,25 @@ def clean(self):
         if self.phone_number == '':
             self.phone_number = None
 
+    @property
+    def permissions(self):
```
nemesisdesign, line 133: let's add a docstring:

```python 
"""
Returns the user permissions from the cache, if the cache is
empty it will call self.get_all_permissions() and cache the result 
"""
```

openwisp_users/base/models.py
```diff
@@ -129,6 +129,25 @@ def clean(self):
...
+        return permissions
+
+    def has_model_perm(self, app_label, model):
+        for perm in self.permissions:
+            if (
+                app_label.lower() == perm.split('.')[0]
+                and model.lower() == perm.split('_')[-1]
+            ):
+                return True
+        return False
```
nemesisdesign, line 149: Why don't we just do:

```
def has_permission(self, permission):
    return permission in self.permissions
```

Since in openwisp-utils, we get the permission like `{app_label}.{model}`.

This method needs a brief mention in the README after `User permissions`

README.rst
```diff
@@ -299,6 +299,12 @@ the token in the ``Authorization`` header:
...
+
+All user permissions are cached and can be accessed as follows ``obj.permissions``. This cache
+is updated each time the user's permissions or group is changed.
+
```
nemesisdesign, line 307:
```
The `User` model of openwisp-users provides a `permissions` helper which returns the user's permissions from the cache, cache invalidation is handled automatically.

.. code-block:: python

    >>> user.permissions
    ... {'account.add_emailaddress',
         'account.change_emailaddress',
         'account.delete_emailaddress',
         'account.view_emailaddress',
         'openwisp_users.add_organizationuser',
         'openwisp_users.add_user',
         'openwisp_users.change_organizationuser',
         'openwisp_users.change_user',
         'openwisp_users.delete_organizationuser',
         'openwisp_users.delete_user'}
```

PR #155, opened 2020-07-26
[admin/minor] User Inlines change form fix

atb00ker:


openwisp_users/admin.py
```diff
@@ -385,16 +385,25 @@ def delete_selected_overridden(self, request, queryset):
...
+                        orguser_readonly = OrganizationUserInlineReadOnly(
+                            self.model, self.admin_site
+                        )
+                        inlines.insert(orguser_index, orguser_readonly)
+                        break
+            return inlines
         inline = OrganizationUserInline(self.model, self.admin_site)
         if request:
-            if hasattr(inline, '_has_add_permission'):
-                has_add_perm = inline._has_add_permission(request, obj)
```
atb00ker, line 395: I looked as far back as Django 1.8 and I don't see `_has_add_permission`. Is this a relic from the far past or am I misunderstanding something?

Moreover, none of the tests seem to fail on removing this. So, I have removed this for asking the same, please let me know if this needs to be added again!

nemesisdesign: Looking at the files in that version, it seems we're not supporting those django versions anymore:
https://github.com/openwisp/openwisp-users/blob/afeac5527599e7efc97706afd0707d33f14bf27c/.travis.yml#L10-L12
Could be removed but I would do that in a separate commit or PR.

PR #159, opened 2020-08-15
[chores] Improved create_organization_owner

nemesisdesign:
- simplified condition: just check if another org owner exists
- catch also IntegrityError (rare, but can happen)

### DIFF
--- a/openwisp_users/apps.py
+++ b/openwisp_users/apps.py
@@ -5,7 +5,7 @@ from django.conf import settings
 from django.contrib.auth import get_user_model
 from django.core.cache import cache
 from django.core.exceptions import ValidationError
-from django.db import transaction
+from django.db import IntegrityError, transaction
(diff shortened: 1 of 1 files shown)


openwisp_users/apps.py
```diff
@@ -100,24 +100,18 @@ def create_organization_owner(cls, instance, created, **kwargs):
...
-            .exclude(pk=instance.pk)
-            .exists()
-        )
-        is_org_owner_exist = OrganizationOwner.objects.filter(
-            organization_user=instance
+        org_owner_exist = OrganizationOwner.objects.filter(
+            organization=instance.organization
         ).exists()
-        if not is_another_org_manager_exist and not is_org_owner_exist:
+        if not org_owner_exist:
```
NoumbissiValere, line 106: :+1: This is alot simpler and less complex. Thanks

PR #160, opened 2020-08-15
[test] Ensure tests fail if inline form details absent

atb00ker:


openwisp_users/tests/test_admin.py
```diff
@@ -31,7 +31,15 @@ class TestUsersAdmin(TestOrganizationMixin, TestUserAdditionalFieldsMixin, TestC
...
+        This function is created to be overridden
+        when the user extends openwisp-users
+        and adds inline forms in the Organization model.
+        """
+        return dict()
+
+    def _get_user_edit_form_inline_params(self, user, organization):
         """
         This function is created to be overridden
         when the user extends openwisp-users
```
atb00ker, line 45: When working on an app that extends `openwisp_users` the organization and user model's inline forms in the extended apps need to be added in the supplied dictonary to run the tests.

This PR:
1. Improves this features and adds it for organization as well.
2. Adds inline forms in `sample_users` to ensure failure of the same is captured.

PR #172, opened 2020-08-24
[refactor] Refactored hard coded tests

pandafy:
Replaced hard coded inline parameters with `add_user_inline_params`


atb00ker (commented):
Thanks @pandafy this was on my todo list, I just couldn't find time for it! :smile:

PR #174, opened 2020-08-27
[refactor] Removed ambiguous test assertions

pandafy:
Removed occurrences where tests were asserting for
"error" string instead of "Please correct the error below."

**EDIT**: I have assumed it is looking for this error message.
Screenshot_20200828_000110


nemesisdesign:
I hope this message won't be changed in django otherwise we'll have to deal with it, but you're right on the fact that the test is not precise enough and may give false positives

PR #175, opened 2020-09-01
[fix] Deleting an OrganizationUser instance may raise exception #165

### LINKED ISSUE #165: [bug] Deleting an OrganizationUser instance may raise exception
Try the following:

- log in as superuser
- create an organization
- create an user for the new organization (will be flagged as owner)
- try deleting the `OrganizationUser` relationship

This exception is raised:

```OwnershipRequired
Cannot delete organization owner before organization or transferring ownership.
```

Expected result:

Instead of an exception being raised, an error should be displayed in the UI: eg `ProtectedError` should do this.

NoumbissiValere:
Fixed issue by ensuring organization user which belongs to an owner
is deleted.

Fixes #165

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -317,6 +317,8 @@ Organization Owners
 An organization owner is a user who is designated as the owner
 of a particular organization and this owner can not be deleted
 or edited by other administrators. Only the superuser has the permissions to do this.
+Also, an error is returned if we try to delete the OrganizationUser instance of an owner
+or flag its ``is_admin`` field as False.
 
 By default, the first manager of an organization is designated as the owner of that organization.
 
--- a/openwisp_users/admin.py
+++ b/openwisp_users/admin.py
@@ -28,6 +28,7 @@ from organizations.base_admin import (
     BaseOrganizationOwnerAdmin,
     BaseOrganizationUserAdmin,
 )
+from organizations.exceptions import OwnershipRequired
 from phonenumber_field.formfields import PhoneNumberField
 from swapper import load_model
 
@@ -436,6 +437,31 @@ class UserAdmin(MultitenantAdminMixin, BaseUserAdmin, BaseAdmin):
                     )
                 )
 
+    def save_formset(self, request, form, formset, change):
+        instances = formset.save(commit=False)
+        not_deleted = 0
+        for obj in formset.deleted_objects:
+            try:
+                obj.delete()
+            except OwnershipRequired:
+                not_deleted += 1
+        if not_deleted:
+            single_msg = (
+                f"Can't delete {not_deleted} organization user because it "
+                'belongs to an organization owner.'
+            )
+            multiple_msg = (
+                f"Can't delete {not_deleted} organization users because they "
+                'belong to some organization owners.'
+            )
+            self.message_user(
+                request,
+                ngettext(single_msg, multiple_msg, not_deleted,),
+                messages.ERROR,
+            )
+        for instance in instances:
+            instance.save()
+
 
 base_fields = list(UserAdmin.fieldsets[1][1]['fields'])
 additional_fields = ['bio', 'url', 'company', 'location', 'phone_number']
@@ -489,6 +515,7 @@ class OrganizationUserAdmin(
     MultitenantAdminMixin, BaseOrganizationUserAdmin, BaseAdmin
 ):
     view_on_site = False
+    actions = ['delete_selected_overridden']
 
     def get_readonly_fields(self, request, obj=None):
         # retrieve readonly fields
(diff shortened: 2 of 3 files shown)


README.rst
```diff
@@ -320,6 +320,19 @@ or edited by other administrators. Only the superuser has the permissions to do
...
+
+Deleting Organization Users
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Becuase of the ``CASCADE`` feature of django, deleting an organization user should
+consequently delete the Organization Owner which depends on it. Doing this may lead
+to confusion as users may delete owners inadvertently. Becuase of this, users are not
+allowed to delete an organization user which belongs to an owner. If they really
+want to do this, then they will have to delete its organization owner first before
+deleting the organization user.
```
nemesisdesign, line 334: I find this confusing. It's not about CASCADE that we're adding this.
I would just keep it simple, is not important, is a corner case, either we should explain it briefly in two lines (but correctly) or remove it.

openwisp_users/admin.py
```diff
@@ -510,6 +512,66 @@ def has_delete_permission(self, request, obj=None):
...
 
+    def delete_view(self, request, object_id, extra_context=None):
+        try:
+            return super().delete_view(request, object_id, extra_context)
+        except OwnershipRequired:
+            self.message_user(
+                request,
+                _(
+                    "Can't delete this organization user because "
+                    "it belongs to an organization owner"
```
nemesisdesign, line 523: use single quotes, applies to all other occurrences

openwisp_users/admin.py
```diff
@@ -510,6 +512,66 @@ def has_delete_permission(self, request, obj=None):
...
+                ),
+                messages.ERROR,
+            )
+            redirect_url = reverse(
+                f'admin:{self.model._meta.app_label}_organizationuser_changelist'
```
nemesisdesign, line 528: redirect to the same page please, not the list

openwisp_users/admin.py
```diff
@@ -510,6 +512,66 @@ def has_delete_permission(self, request, obj=None):
...
+        actions = super().get_actions(request)
+        if not request.POST.get('post') and 'delete_selected' in actions:
+            del actions['delete_selected']
+        return actions
+
+    def delete_selected_overridden(self, request, queryset):
+        count = 0
+        pks = []
+        for obj in queryset:
+            if getattr(obj, 'organizationowner', None):
```
nemesisdesign, line 542: use the membership helpers...

openwisp_users/admin.py
```diff
@@ -510,6 +512,67 @@ def has_delete_permission(self, request, obj=None):
...
 
+    def delete_view(self, request, object_id, extra_context=None):
+        try:
+            return super().delete_view(request, object_id, extra_context)
+        except OwnershipRequired:
+            self.message_user(
+                request,
+                _(
+                    "Can't delete this organization user because "
+                    'it belongs to an organization owner'
```
nemesisdesign, line 523: add a final dot please

openwisp_users/admin.py
```diff
@@ -510,6 +512,67 @@ def has_delete_permission(self, request, obj=None):
...
+        actions = super().get_actions(request)
+        if not request.POST.get('post') and 'delete_selected' in actions:
+            del actions['delete_selected']
+        return actions
+
+    def delete_selected_overridden(self, request, queryset):
+        count = 0
+        pks = []
+        for obj in queryset:
+            if obj.user.is_owner(obj.organization_id):
```
nemesisdesign, line 543: please add an `assertNumQueries` on the tests which trigger this line: to me it looks it could generate too many queries.
If it does generate too many queries, `select_related` has to be used.

README.rst
```diff
@@ -317,6 +317,8 @@ Organization Owners
 An organization owner is a user who is designated as the owner
 of a particular organization and this owner can not be deleted
 or edited by other administrators. Only the superuser has the permissions to do this.
+Also, an error is returned if we try to delete the OrganizationUser instance of an owner
+or flag its ``is_admin`` field as False.
```
nemesisdesign, line 321: note for myself: this needs further improvement

openwisp_users/tests/test_admin.py
```diff
@@ -1275,6 +1275,76 @@ def test_only_superuser_can_delete_inline_org_owner(self):
...
+            self.assertEqual(qs.count(), 0)
+
+        with self.subTest("Can not delete only owner's org user with action"):
+            post_data = {
+                'action': 'delete_selected_overridden',
+                '_selected_action': [org_user.pk],
+                'post': 'yes',
+            }
+            url = reverse(f'admin:{self.app_label}_organizationuser_changelist')
+            r = self.client.post(url, post_data, follow=True)
```
NoumbissiValere, line 1319: This is the point which calls `delete_selected_overridden`. with the `delete_selected_overridden`, the number of queries this line shows is `8` and when `delete_selected_overridden` is commented out (that is using the default `delete_selected`) the number of queries returned is `23`.

openwisp_users/tests/test_admin.py
```diff
@@ -1275,6 +1275,76 @@ def test_only_superuser_can_delete_inline_org_owner(self):
...
+            self.assertEqual(r.status_code, 200)
+            msg = (
+                't delete 1 organization user because '
+                'it belongs to an organization owner'
+            )
+            self.assertContains(r, msg)
+            post_data.update({'post': 'yes'})
+            r = self.client.post(url, post_data, follow=True)
```
NoumbissiValere, line 1341: This is the point which calls `delete_selected_overridden`. with the `delete_selected_overridden`, the number of queries this line shows is `17` and when `delete_selected_overridden` is commented out (that is using the default delete_selected) the number of queries returned is `16`. and the extra query can be accounted for by the line which does `queryset = queryset.exlcude(pk_in=pks)`

openwisp_users/admin.py
NoumbissiValere: @nemesisdesign i don't think it is in this PR that we should be checking if the `helpers` produce extra queries or not. I believe that should have been done in the PR which had that solution. But from the comments below, it shows no extra queries are executed here.

nemesisdesign: the test does not have the aim of testing the helper, we already have a test for that, it has the aim to test this method, which may be changed by somebody else in the future and hence may break it or modify in a way that generates too many queries, so please do add it, as I already explained we need to start to keep in check the number of queries our code generates because it's getting out of hand. We already started this process: https://github.com/search?q=org%3Aopenwisp+assertNumQueries&type=Code
We have to stick at it and that's why I'm insisting with you. Please make it a habit, it's a very good one.

openwisp_users/tests/test_admin.py
nemesisdesign: cool, please add the `assertNumQueries` both here and above.
If there's any discrepancies due to the django version used, you can follow this approach: https://github.com/openwisp/openwisp-network-topology/blob/dff2357602d24f24b71e3e7aa83d61e8b94718af/openwisp_network_topology/integrations/device/tests.py#L93-L96

nemesisdesign (changes requested):
Good progress.

Look at this: https://github.com/openwisp/openwisp-users/blob/master/openwisp_users/settings.py#L4
The organization user admin is disabled by default, is generally not used in prod, although it may, so your patch fixes part of the problem and that's good.

But it does not fix the main problem: try deleting an organizationuser relationship from the edit user details page as in the gif below:

openwisp-users-issue-165

See also my comments below.

Once these are fixed this will be ready to be merged.

nemesisdesign (dismissed):
Good work @NoumbissiValere, I will add a couple minor improvements asap and then merge.

openwisp_users/tests/test_admin.py
```diff
@@ -1275,6 +1275,78 @@ def test_only_superuser_can_delete_inline_org_owner(self):
...
+            self.assertEqual(qs.count(), 0)
+
+        with self.subTest("Can not delete only owner's org user with action"):
+            post_data = {
+                'action': 'delete_selected_overridden',
+                '_selected_action': [org_user.pk],
+                'post': 'yes',
+            }
+            url = reverse(f'admin:{self.app_label}_organizationuser_changelist')
+            with self.assertNumQueries(8):
```
nemesisdesign, line 1319: note for self: this fails in the integration test, I have to find a solution to disable this assertion in integration tests with other modules

nemesisdesign (approved):

PR #176, opened 2020-09-01
[fix] OrganizationOwner shouldn't be allowed to be is_admin=False #166

### LINKED ISSUE #166: [bug] OrganizationOwner shouldn't be allowed to be is_admin=False
Create an organization user which becomes owner of an organization.

Try to flag it as OrganizationUser(is_admin=False).

The system allows to do this but it shouldn't, it's confusing, because the user will keep being recognized as owner.

The correct behavior would be to raise a ValidationError which says the user is owner of this organization and cannot be downgraded.

NoumbissiValere:
Fixed issue by preventing org owners from been downgraded.

Fixes #166

### DIFF
--- a/openwisp_users/base/models.py
+++ b/openwisp_users/base/models.py
@@ -224,6 +224,15 @@ class BaseOrganizationUser(models.Model):
     class Meta:
         abstract = True
 
+    def clean(self):
+        if self.user.is_owner(str(self.organization_id)) and not self.is_admin:
+            raise ValidationError(
+                _(
+                    f'{self.user.username} is the owner of the organization: '
+                    f'{self.organization}, and cannot be downgraded'
+                )
+            )
+
 
 class BaseOrganizationOwner(models.Model):
     """
--- a/openwisp_users/base/models.py
+++ b/openwisp_users/base/models.py
@@ -83,8 +83,12 @@ class AbstractUser(BaseUser):
         """ meant for internal usage only """
         if isinstance(obj, (uuid.UUID, str)):
             pk = obj
-        else:
+        elif isinstance(obj, BaseOrganization):
             pk = obj.pk
+        elif obj is None:
+            return None
+        else:
+            raise ValueError('expected UUID, str or Organization instance')
         return str(pk)
 
     def is_member(self, organization):
@@ -225,7 +229,7 @@ class BaseOrganizationUser(models.Model):
         abstract = True
 
     def clean(self):
-        if self.user.is_owner(str(self.organization_id)) and not self.is_admin:
+        if self.user.is_owner(self.organization_id) and not self.is_admin:
             raise ValidationError(
                 _(
                     f'{self.user.username} is the owner of the organization: '
--- a/openwisp_users/tests/test_models.py
+++ b/openwisp_users/tests/test_models.py
@@ -289,3 +289,26 @@ class TestUsers(TestOrganizationMixin, TestCase):
         self.assertEqual(User._get_pk(org), pk)
         self.assertEqual(User._get_pk(org.pk), pk)
         self.assertEqual(User._get_pk(pk), pk)
+
+    def test_orguser_is_admin_change(self):
+        org = self._create_org(name='test-org')
+        user1 = self._create_user(username='user1', email='user1@email.com')
+        user2 = self._create_user(username='user2', email='user2@email.com')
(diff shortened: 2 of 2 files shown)


openwisp_users/tests/test_models.py
```diff
@@ -263,3 +263,25 @@ def test_superuser_has_permission(self):
...
+        org_user2 = self._create_org_user(user=user2, organization=org, is_admin=True)
+
+        with self.subTest('change is_admin when org_user belongs to org_owner'):
+            msg = (
+                f'{user1.username} is the owner of the organization: '
+                f'{org}, and cannot be downgraded'
+            )
+            with self.assertRaisesMessage(ValidationError, msg):
+                org_user1.is_admin = False
+                org_user1.full_clean()
```
nemesisdesign, line 281: could you please also use [assertNumQueries](https://docs.djangoproject.com/en/3.1/topics/testing/tools/#django.test.TransactionTestCase.assertNumQueries) to find out how many queries are executed here?

openwisp_users/base/models.py
```diff
@@ -215,6 +215,15 @@ class BaseOrganizationUser(models.Model):
     class Meta:
         abstract = True
 
+    def clean(self):
+        if getattr(self, 'organizationowner', None) and not self.is_admin:
```
nemesisdesign, line 219: why not use the helpers method we created? We should avoid queries as much as possible:

```python
if self.user.is_owner(self.organization) and not self.is_admin:
    ...
```

or alternatively:

```python
if str(self.organization.pk) in self.user.organizations_owned and not self.is_admin:
    ...
```

See my comment below about `assertNumQueries`, can you quickly expriment with the 3 approaches and let me know which one generates less queries?

NoumbissiValere: @nemesisdesign all 3 cases use just one query. I think `getattr(self, 'organizationowner', None)` didn't generate extra queries because it was gotten using `select_related`. But we can't and shouldn't depend on that so I am using `self.user.is_owner(self.organization)`

openwisp_users/base/models.py
```diff
@@ -224,6 +224,15 @@ class BaseOrganizationUser(models.Model):
     class Meta:
         abstract = True
 
+    def clean(self):
+        if self.user.is_owner(str(self.organization_id)) and not self.is_admin:
```
NoumbissiValere, line 228: `str(self.organization_id)` is just to ensure that if `self.organization_id` is `None` [obj.pk](https://github.com/openwisp/openwisp-users/blob/master/openwisp_users/base/models.py#L87) should not raise a `NoneType does not have pk` error

nemesisdesign: This is not needed, see here: https://github.com/openwisp/openwisp-users/commit/85dfe0c1dc7c61926cb1e3b0888ecadeb22b4398#diff-4e7304732af6cee239b0aeb34f221e7eR81-R98
Just rebase on the current master and it will be fine.

NoumbissiValere: @nemesisdesign I rebased on master before being able to do `self.user.is_owner(str(self.organization_id))`. remember the old `is_owner` only accepted an organization instance but you see here, i am sending the `organization_id` instead.

But, because of [this test](https://github.com/openwisp/openwisp-users/blob/master/openwisp_users/tests/test_admin.py#L642), the `clean` method is called with `self.organization_id=None` since no organization is provided (which is correct as that is the objective of the test). when this happens, if i use `user.is_owner(self.organization_id)` `None` will be passed down and since `None` is neither a `UUID` nor `str` instance, [this line](https://github.com/openwisp/openwisp-users/commit/85dfe0c1dc7c61926cb1e3b0888ecadeb22b4398#diff-4e7304732af6cee239b0aeb34f221e7eR84) fails and control moves on to [this line](https://github.com/openwisp/openwisp-users/commit/85dfe0c1dc7c61926cb1e3b0888ecadeb22b4398#diff-4e7304732af6cee239b0aeb34f221e7eR87) and will try to do `pk=None.pk` which will raise the error `NoneType does not have the attribute pk`. 

I did  `user.is_owner(str(self.organization_id))`, because if `self.organization_id=None`, the string version (`"None"`) is passed down and no error will be encountered and the correct value will be returned.

nemesisdesign (approved):

PR #179, opened 2020-10-01
[chores] Fixed typographical error in organization name #173

### LINKED ISSUE #173: [chores] Typographical error in OPENWISP_ORGANIZATON_* setting
Fix typographical error in `OPENWISP_ORGANIZATON_USER_ADMIN` and `OPENWISP_ORGANIZATON_OWNER_ADMIN` settings.
It should be `OPENWISP_ORGANIZATION_*` instead. 
Fix all occurrence of these settings (docs and code). 

These settings are also used in other modules. After fixing it here, a patch should be sent to all those repositories.

Following is a non-exhaustible list of modules containing such occurrences:

- [ ] [openwisp-ipam](https://github.com/openwisp/openwisp-ipam/blob/master/tests/openwisp2/settings.py#L86-L87)
- [ ] [openwisp-network-topology](https://github.com/openwisp/openwisp-network-topology/blob/master/tests/openwisp2/settings.py#L147-L148)
- [ ] [openwisp-radius](https://github.com/openwisp/openwisp-radius/blob/master/tests/openwisp2/settings.py#L227-L228) 
- [ ] [openwisp-controller](https://github.com/openwisp/openwisp-controller/blob/master/tests/openwisp2/settings.py#L125)

devkapilbansal:
Closes #173

### DIFF
--- a/openwisp_users/settings.py
+++ b/openwisp_users/settings.py
@@ -4,14 +4,14 @@ from openwisp_utils.utils import default_or_test
 ORGANIZATION_USER_ADMIN = (
     # setting with typographic error for maintaining backward compatibility
     # TODO: remove in openwisp-users 0.5
-    getattr(settings, 'OPENWISP_ORGANIZATON_USER_ADMIN', False) or
-    getattr(settings, 'OPENWISP_ORGANIZATION_USER_ADMIN', False)
+    getattr(settings, 'OPENWISP_ORGANIZATON_USER_ADMIN', False)
+    or getattr(settings, 'OPENWISP_ORGANIZATION_USER_ADMIN', False)
 )
 ORGANIZATION_OWNER_ADMIN = (
     # setting with typographic error for maintaining backward compatibility
     # TODO: remove in openwisp-users 0.5
-    getattr(settings, 'OPENWISP_ORGANIZATON_OWNER_ADMIN', True) or
     getattr(settings, 'OPENWISP_ORGANIZATON_OWNER_ADMIN', True)
+    or getattr(settings, 'OPENWISP_ORGANIZATON_OWNER_ADMIN', True)
 )
 USERS_AUTH_API = getattr(settings, 'OPENWISP_USERS_AUTH_API', False)
 USERS_AUTH_THROTTLE_RATE = getattr(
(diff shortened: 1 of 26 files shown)


devkapilbansal:
Will soon send a patch to other repositories as well.

openwisp_users/settings.py
```diff
@@ -1,8 +1,8 @@
 from django.conf import settings
 from openwisp_utils.utils import default_or_test
 
-ORGANIZATON_USER_ADMIN = getattr(settings, 'OPENWISP_ORGANIZATON_USER_ADMIN', False)
-ORGANIZATON_OWNER_ADMIN = getattr(settings, 'OPENWISP_ORGANIZATON_OWNER_ADMIN', True)
+ORGANIZATION_USER_ADMIN = getattr(settings, 'OPENWISP_ORGANIZATION_USER_ADMIN', False)
```
nemesisdesign, line 4: unfortunately we need to be backward compatible with the mistaken setting, otherwise existing applications will change inadvertently. 

something like this should work:

```python
ORGANIZATION_USER_ADMIN = (
    # setting with typographic error maintained for backward compatibility
    # TODO: remove in openwisp-users 0.5
    getattr(settings, 'OPENWISP_ORGANIZATON_USER_ADMIN', False) or
    getattr(settings, 'OPENWISP_ORGANIZATION_USER_ADMIN', False)
)
```

We'll then create an issue to remind us to remove this in the next point version.

nemesisdesign (changes requested):
Good progress @devkapilbansal, see my comment below.

openwisp_users/settings.py
devkapilbansal: Okay, will do that. Learned something new today :smile:

devkapilbansal:
@nemesisdesign in the latest commit, I solve some RST errors in README file too. So, not sure if the type should be **refactor**, **chores** or **docs**

openwisp_users/settings.py
devkapilbansal: @nemesisdesign Changing 'ORGANIZATION' back to 'ORGANIZATON' results in test failure. Maybe we should keep it on hold until a new version is released.

nemesisdesign (approved):
Thanks!

PS: please if you can help in updating the other repos it would be very helpful

PR #192, opened 2020-11-09
[fix] Allow swagger to show parameters of obtain token view #116

### LINKED ISSUE #116: [api] Allow swagger to show parameters of obtain token view
Allow swagger to show parameters of obtain token view.
Right now when testing the obtain token view via the swagger API docs there's no way to specify username and password.

niteshsinha17:
closes #116

### DIFF
--- a/openwisp_users/api/swagger.py
+++ b/openwisp_users/api/swagger.py
@@ -1,0 +1,16 @@
+"""
+The following serializers are not used to send
+response to user but in swagger documentation of
+the API.
+"""
+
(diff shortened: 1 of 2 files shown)


nemesisdesign (changes requested):
Good progress, see my comments below

niteshsinha17:
@nemesisdesign I have made changes as you had suggested. If I have missed something then please let me know.

nemesisdesign (approved):
Thanks :+1:

PR #193, opened 2020-11-12
[fix] Select automatically organization when only 1 org #170

### LINKED ISSUE #170: [admin] Make organization selection automatic when only 1 org
When the user logged in has access to only one organization, the organization select should automatically set that org as the default value, the user should not need to select it.

niteshsinha17:
closes #170

### DIFF
--- a/openwisp_users/admin.py
+++ b/openwisp_users/admin.py
@@ -105,9 +105,6 @@ class OrganizationUserInline(admin.StackedInline, SelectOrgMixin):
         display all organizations
         """
         formset = super().get_formset(request, obj=obj, **kwargs)
-        if request.user.is_superuser:
-            self.select_organization(request, formset.form)
-            return formset
         if not request.user.is_superuser:
             formset.form.base_fields[
                 'organization'
--- a/openwisp_users/admin.py
+++ b/openwisp_users/admin.py
@@ -33,7 +33,7 @@ from phonenumber_field.formfields import PhoneNumberField
 from swapper import load_model
 
 from . import settings as app_settings
-from .multitenancy import MultitenantAdminMixin
+from .multitenancy import MultitenantAdminMixin, SelectOrgMixin
 from .utils import BaseAdmin
 
 Group = load_model('openwisp_users', 'Group')
@@ -92,7 +92,7 @@ class OrganizationOwnerInline(admin.StackedInline):
         return super().has_change_permission(request, obj)
 
 
-class OrganizationUserInline(admin.StackedInline):
+class OrganizationUserInline(admin.StackedInline, SelectOrgMixin):
     model = OrganizationUser
     formset = RequiredInlineFormSet
     view_on_site = False
@@ -106,6 +106,7 @@ class OrganizationUserInline(admin.StackedInline):
         """
         formset = super().get_formset(request, obj=obj, **kwargs)
         if request.user.is_superuser:
+            self.select_organization(request, formset.form)
             return formset
         if not request.user.is_superuser:
             formset.form.base_fields[
@@ -113,6 +114,7 @@ class OrganizationUserInline(admin.StackedInline):
             ].queryset = Organization.objects.filter(
                 pk__in=request.user.organizations_managed
             )
+        self.select_organization(request, formset.form)
         return formset
 
     def get_extra(self, request, obj=None, **kwargs):
--- a/openwisp_users/apps.py
+++ b/openwisp_users/apps.py
@@ -54,6 +54,7 @@ class OpenwispUsersConfig(AppConfig):
     def connect_receivers(self):
         OrganizationUser = load_model('openwisp_users', 'OrganizationUser')
         OrganizationOwner = load_model('openwisp_users', 'OrganizationOwner')
+        Organization = load_model('openwisp_users', 'Organization')
         Group = load_model('openwisp_users', 'Group')
         User = get_user_model()
         signal_tuples = [(post_save, 'post_save'), (post_delete, 'post_delete')]
@@ -67,6 +68,12 @@ class OpenwispUsersConfig(AppConfig):
                         name, model.__name__
                     ),
                 )
+        for signal, name in signal_tuples:
+            signal.connect(
+                self.update_selected_org,
+                sender=Organization,
+                dispatch_uid='update_selected_org',
+            )
         post_save.connect(
             self.create_organization_owner,
             sender=OrganizationUser,
@@ -82,6 +89,15 @@ class OpenwispUsersConfig(AppConfig):
                 sender=model,
                 dispatch_uid='update_user_permissions',
             )
+
+    def update_selected_org(cls, user=None, **kwargs):
+        if user:
+            cache_key = 'selected_{}_org'.format(user.pk)
+            cache.delete(cache_key)
+
+        # also delete for superuser
+        cache_key = 'selected_{}_org'.format('superuser')
+        cache.delete(cache_key)
 
     def update_organizations_dict(cls, instance, **kwargs):
         if hasattr(instance, 'user'):
@@ -100,6 +116,7 @@ class OpenwispUsersConfig(AppConfig):
             del user.organizations_owned
         except AttributeError:
             pass
+        cls.update_selected_org(user, **kwargs)
 
     def create_organization_owner(cls, instance, created, **kwargs):
         if not created or not instance.is_admin:
--- a/openwisp_users/apps.py
+++ b/openwisp_users/apps.py
@@ -95,11 +95,11 @@ class OpenwispUsersConfig(AppConfig):
 
     def update_selected_org(cls, user=None, **kwargs):
         if user:
-            cache_key = 'selected_{}_org'.format(user.pk)
+            cache_key = f'selected_{user.pk}_org'
             cache.delete(cache_key)
 
         # also delete for superuser
-        cache_key = 'selected_{}_org'.format('superuser')
+        cache_key = 'selected_superuser_org'
         cache.delete(cache_key)
 
     def update_organizations_dict(cls, instance, **kwargs):
@@ -107,7 +107,7 @@ class OpenwispUsersConfig(AppConfig):
             user = instance.user
         else:
             user = instance.organization_user.user
-        cache_key = 'user_{}_organizations'.format(user.pk)
+        cache_key = f'user_{user.pk}_organizations'
         cache.delete(cache_key)
         # forces caching
         user.organizations_dict
--- a/openwisp_users/apps.py
+++ b/openwisp_users/apps.py
@@ -68,12 +68,13 @@ class OpenwispUsersConfig(AppConfig):
                         name, model.__name__
                     ),
                 )
-        for signal, name in signal_tuples:
-            signal.connect(
-                self.update_selected_org,
-                sender=Organization,
-                dispatch_uid='update_selected_org',
-            )
+        for model in [Organization]:
+            for signal, name in signal_tuples:
+                signal.connect(
+                    self.update_selected_org,
+                    sender=Organization,
+                    dispatch_uid='update_selected_org',
+                )
         post_save.connect(
             self.create_organization_owner,
             sender=OrganizationUser,
--- a/openwisp_users/apps.py
+++ b/openwisp_users/apps.py
@@ -72,8 +72,10 @@ class OpenwispUsersConfig(AppConfig):
             for signal, name in signal_tuples:
                 signal.connect(
                     self.update_selected_org,
-                    sender=Organization,
-                    dispatch_uid='update_selected_org',
+                    sender=model,
+                    dispatch_uid='{}_{}_update_selected_org'.format(
+                        name, model.__name__
+                    ),
                 )
         post_save.connect(
             self.create_organization_owner,
--- a/openwisp_users/base/models.py
+++ b/openwisp_users/base/models.py
@@ -113,6 +113,7 @@ class AbstractUser(BaseUser):
         else:
             selected_org = False  # Tells selected org is None
         cache.set(cache_key, selected_org, 86400 * 2)  # Cache for two days
+        return selected_org
 
     @property
     def organizations_dict(self):
--- a/openwisp_users/base/models.py
+++ b/openwisp_users/base/models.py
@@ -97,6 +97,23 @@ class AbstractUser(BaseUser):
                 return True
         return False
 
+    def selected_org(self, org_field):
+        if self.is_superuser:
+            cache_key = 'selected_{}_org'.format('superuser')
+        else:
+            cache_key = 'selected_{}_org'.format(self.pk)
+        selected_org = cache.get(cache_key)
+        if selected_org is not None:  # If cache is not deleted
+            if selected_org is False:
+                return None
+            return selected_org
+        org_queryset = org_field._queryset
+        if org_queryset.count() == 1:
+            selected_org = org_queryset[0]
+        else:
+            selected_org = False  # Tells selected org is None
+        cache.set(cache_key, selected_org, 86400 * 2)  # Cache for two days
+
     @property
     def organizations_dict(self):
         """
--- a/openwisp_users/base/models.py
+++ b/openwisp_users/base/models.py
@@ -99,9 +99,9 @@ class AbstractUser(BaseUser):
 
     def selected_org(self, org_field):
         if self.is_superuser:
-            cache_key = 'selected_{}_org'.format('superuser')
+            cache_key = 'selected_superuser_org'
         else:
-            cache_key = 'selected_{}_org'.format(self.pk)
+            cache_key = f'selected_{self.pk}_org'
         selected_org = cache.get(cache_key)
         if selected_org is not None:  # If cache is not deleted
             if selected_org is False:
@@ -113,6 +113,8 @@ class AbstractUser(BaseUser):
         else:
             selected_org = False  # Tells selected org is None
         cache.set(cache_key, selected_org, 86400 * 2)  # Cache for two days
+        if selected_org is False:
+            return None
         return selected_org
 
     @property
--- a/openwisp_users/multitenancy.py
+++ b/openwisp_users/multitenancy.py
@@ -8,7 +8,21 @@ User = get_user_model()
 OrganizationUser = load_model('openwisp_users', 'OrganizationUser')
 
 
-class MultitenantAdminMixin(object):
+class SelectOrgMixin(object):
+    def select_organization(self, request, form):
+        '''
+        When the user has access to only one organization
+        automatically set that org as the default value
+        '''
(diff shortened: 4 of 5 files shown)


dwang:
Works for me. Tested with superuser and non-superuser.

NoumbissiValere (approved):
Great job @niteshsinha17 I have tested this and it works well. Leaving for @nemesisdesign to validate and merge

openwisp_users/multitenancy.py
```diff
@@ -8,7 +8,23 @@
...
 
 
-class MultitenantAdminMixin(object):
+class SelectOrgMixin(object):
+    def select_organization(self, request, form):
+        '''
+        When the user has access to only one organization
+        automatically set that org as the default value
+        '''
+        if request.POST:
```
NoumbissiValere, line 17: @niteshsinha17 if we are checking if the method of the `request` is `POST`, i will suggest we use `if request.method == 'POST'` instead of `if request.POST`. 

Also, if this is the case, we should not send the entire `request` object since it is only use to check the `request method`, instead, we could send the `request.method` as a parameter to the `select_organization` method

niteshsinha17: It sounds good. I will change it.

niteshsinha17: But I think check should be done inside select_organization method, it will prevent writing check line multiple times in future if this method is used somewhere else. Or if in future if we want to do something with request in the select_organization function then we will need to change check everywhere where this function is being used. I think this is not good of coding. What do you think?

NoumbissiValere: @niteshsinha17 what I meant was that `select_organization` should be called like this: `selecte_organization(request.method, form)` and in the `select_organization` method definition, you can do the check with `if method == 'POST'`. But It looks cleaner to just send `request` to the method. But please check for the method the way i describe. you can read more about why from [django docs](https://docs.djangoproject.com/en/3.1/ref/request-response/#django.http.HttpRequest.POST). 


So in conclusion, you could leave the function call the way it is now but in your method definition, you should do;
```
def select_organization(request, form):
    if request.method == 'POST':
        return
   ...
```

NoumbissiValere (changes requested):
@niteshsinha17 Please add a test case for this fix you provided so that someone will not mistakenly remove this great work of yours in the future.

hint:
in your test case,
- make sure there is only one organization in the database
- visit the add page of `User` and verify that you can see this: `selected>org_name</option>`

test with a superuser and non-superuser who is an organization owner.

nemesisdesign (changes requested):
I confirm this needs tests, thanks @dwang for testing it, thanks @NoumbissiValere for reviewing and mentioning the need for the test.

openwisp_users/multitenancy.py
niteshsinha17: I agree with your points. 😃

niteshsinha17:
@NoumbissiValere I don't have any idea of writing tests. But i will try to learn and understand it through the code after that i will add tests for it as soon as possible.

NoumbissiValere:
> @NoumbissiValere I don't have any idea of writing tests. But i will try to learn and understand it through the code after that i will add tests for it as soon as possible.

@niteshsinha17 you could also get help from [django testing docs](https://docs.djangoproject.com/en/3.1/topics/testing/)

openwisp_users/multitenancy.py
```diff
@@ -8,7 +8,23 @@
...
+            return
+        fields = form.base_fields
+        if 'organization' in fields:
+            org_field = fields['organization']
+            org_queryset = org_field._queryset
+            if org_queryset.count() == 1:
+                org_field.initial = org_queryset[0]
```
nemesisdesign, line 24: There's two problems here.

1. The system may have multiple organizations but the user has only access to one.
2. The query will be repeated in every page, but will yield always the same result for a given user

For 1, I'm not sure this code addresses this. Does it?
We surely need a test for this.

Alternatively to check that a nonsuper user has access to only 1 org we could use:
```python
if not request.user.is_superuser and len(request.user.organization_dict.keys()) == 1:
    # ...
```

But only I'd use it only if this code does not already solve it, it may be possible that the queryset is already preprocessed by the multitenancy admin mixin, would be good if you confirm this.

For problem 2., I think we should cache the query for each user and add caching invalidation, like we already do for organizations_dict:

- caching: https://github.com/openwisp/openwisp-users/blob/3e28715c94617faeab5d23514dfa3ccbbf217975/openwisp_users/base/models.py#L94-L120
- invalidation: https://github.com/openwisp/openwisp-users/blob/master/openwisp_users/apps.py#L86-L102

nemesisdesign (changes requested):
For the tests, I suggest taking a look at this test suite: https://github.com/openwisp/openwisp-users/blob/master/openwisp_users/tests/test_admin.py, as well as to study the django documentation which is very detailed.

See also my comment below.

niteshsinha17:
@nemesisdesign  I have confirmed queryset is already preprocessed by the multitenancy admin mixin.

openwisp_users/multitenancy.py
niteshsinha17: In the invalidation when `update_organizations_dict` function is called?

niteshsinha17: I have confirmed the queryset is already preprocessed by the multitenancy admin mixin and the code is solving 1st problem.

niteshsinha17: @nemesisdesign  I have tried to cached the query using following code.

```
def select_organization(self, request, form):
        '''
        When the user has access to only one organization
        automatically set that org as the default value
        '''
        if request.POST:
            return
        fields = form.base_fields
        if 'organization' in fields:
            org_field = fields['organization']
            org_field.initial = request.user.selected_org(org_field)

```
and 
```
def selected_org(self,org_field):
        # print('here')
        # print(self.username)
        cache_key = 'selected_{}_org'.format(self.pk)
        selected_org = cache.get(cache_key)
        # print(selected_org)
        if selected_org is not None: # If cache is not deleted
            # print('cached')
            if selected_org == False:
                return None 
            return selected_org
        # print('not cached')
        org_queryset = org_field._queryset
        if org_queryset.count() == 1:
                selected_org = org_queryset[0]
        else:
            selected_org = False  # Tells selected org is None
        cache.set(cache_key, selected_org, 86400 * 2)  # Cache for two days
``` 
This in models.py of base.

But I am able to delete the cache with the help of invalidation as you have suggested because I can't get request.user in signal.

niteshsinha17: I have added caching. Please review it once so that I can work on testing.

openwisp_users/base/models.py
```diff
@@ -109,6 +109,23 @@ def is_owner_of_any_organization(self):
...
+        selected_org = cache.get(cache_key)
+        if selected_org is not None:  # If cache is not deleted
+            if selected_org is False:
+                return None
+            return selected_org
+        org_queryset = org_field._queryset
+        if org_queryset.count() == 1:
+            selected_org = org_queryset[0]
+        else:
+            selected_org = False  # Tells selected org is None
```
NoumbissiValere, line 126: While not set `selected_org=None`

openwisp_users/base/models.py
```diff
@@ -109,6 +109,23 @@ def is_owner_of_any_organization(self):
...
         return False
 
+    def selected_org(self, org_field):
+        if self.is_superuser:
+            cache_key = 'selected_{}_org'.format('superuser')
+        else:
+            cache_key = 'selected_{}_org'.format(self.pk)
+        selected_org = cache.get(cache_key)
+        if selected_org is not None:  # If cache is not deleted
+            if selected_org is False:
```
NoumbissiValere, line 119: when there's no org to select, set `selected_org` to None so that this second if while not be needed

niteshsinha17: If you will notice carefully then you will realise that if we set it to None then query will be evaluated for None. I want this function to return None or an org. That's why I am storing False in place of None which will indicate it have to return None and query will not be evaluated for None always.

NoumbissiValere: @nemesisdesign is this cache at the right place? It makes more sense to me if it was on the `organization model`

niteshsinha17: I am making cache key differently for superuser because of following reason. 

1. When an org is deleted then Org Owner and Org User is also deleted which sends a signal that is received in `apps.py` file. There in instance we get user who is associated with the deleted org but superuser is not associated with any org. So we can't delete cache stored for superuser. 

I preferred to add cache here because we need to have cache for every user.

niteshsinha17:
I have added test for it. Please review it once again.

nemesisdesign (commented):
@niteshsinha17 sorry I forgot to add my review last time I tested this, I didn't see the organization being selected automatically when the user had only one org, could you share a gif of the result you get please?

niteshsinha17:
> @niteshsinha17 sorry I forgot to add my review last time I tested this, I didn't see the organization being selected automatically when the user had only one org, could you share a gif of the result you get please?

Sure, I have shared it.

org_select

I think cache is store in your system. You may need to create or delete an organization to change it.

nemesisdesign (changes requested):
First of all, great work @niteshsinha17!
It works, almost. Please read on.

I've tested it again and found the issue.

Here's how to replicate it.

- launch the dev server
- log in as admin, make sure only one org is present in the system
- go to the add a new book page
- stop the dev server and relaunch
- reload the page, no organization is selected

However, after spending time to test your work in this PR and @Purhan's work in https://github.com/openwisp/openwisp-users/pull/209, I realized the following:

Operators of specific organizations cannot create shared objects, so the empty label gets removed from their choices and only the organization they manage will be shown as default option, which solves the issue you're working on for operators.

Superusers, when the object can be shared between multiple organizations, will see a label indicating this, which is a good default option.

The last case is when superusers add a new object which is not shareable and there's only 1 org in the system, it would be nice to auto-select the organization in this case but I'm not sure if the all the lines of code in this solution are worth it.

Maybe we could try simplifying the solution.

What about editing the `_edit_form` method like it has been done in https://github.com/openwisp/openwisp-users/pull/209/files#diff-a854d6df2d833169143f34631ee4c6c4f8bd858377613c8ac309dea3be7843a6L68-R106, we could do the following:

- check if user is superuser and org field is required and only 1 org in the system, in this case also remove the empty label (which forces the selection of the first org)
- the query to check if there's only 1 org can be cached, there would be only one cache key (will be used only by superusers)
- the invalidation should be done by listening to add/delete signals coming from the organization model like we do here for other models: https://github.com/openwisp/openwisp-users/blob/69cd57a122243ad53b79629c401df584880605b9/openwisp_users/apps.py#L54-L84  

Here's an example:

```python
def _edit_form(self, request, form):
        """
        Modifies the form querysets as follows;
        if current user is not superuser:
            * show only relevant organizations
            * show only relations associated to relevant organizations
              or shared relations
        else show everything
        """
        fields = form.base_fields
        user = request.user
        org_field = fields.get('organization')
        orgs_pk = user.organizations_managed
        if user.is_superuser and org_field:
            if not org_field.required:
                org_field.empty_label = _('Shared systemwide (no organization)')
            elif self._get_count(org_field.queryset) == 1:
                org_field.empty_label = None

def _get_count(self, queryset):
    # get count from cache if present
    # otherwise call org_field.queryset.count()
```

This would radically simplify the solution and reduce the amount of lines required. What do you think?

Please also rebase on the latest master.

niteshsinha17:
@nemesisdesign Thanks for the review.
> I've tested it again and found the issue.

I tried to replicate this but it was not occurring in my system. During this, I found one more Issue and I have solved it. I think it may have also solved the issue about which you are talking. 

> This would radically simplify the solution and reduce the amount of lines required. What do you think?
I think we need to do the same thing during the creation of a user and it's not going to work for it. 
We can modify `SelectOrgMixin`  to simplify our work for both cases. 
```
class SelectOrgMixin(object):
    def select_organization(self, request, form):
        '''
        When the user has access to only one organization
        automatically set that org as the default value
        '''
        if request.method == 'POST':
            return
        fields = form.base_fields
        if 'organization' in fields:
            org_field = fields['organization']
            org_field.initial = request.user.selected_org(org_field)
            org_field.empty_label = self._get_empty_level(
                request.user.is_superuser, org_field
            )

    def _get_empty_level(self, superuser, org_field):
        if superuser:
            if not org_field.required:
                return _('Shared systemwide (no organization)')
            elif org_field._queryset.count() == 1:
                return None
            return org_field.empty_label
        return None
```
Then we can remove these lines too.
https://github.com/openwisp/openwisp-users/blob/a23e05fee407f739803740a0e800fc1fa3ad33f4/openwisp_users/multitenancy.py#L70-L71
and
https://github.com/openwisp/openwisp-users/blob/a23e05fee407f739803740a0e800fc1fa3ad33f4/openwisp_users/multitenancy.py#L78
What do you think?

Also,
When object can be shared between multiple organizations and only one organization is present then which should be selected as default?
This case is happening with the `Template` model and tests are failing at these lines. 

https://github.com/openwisp/openwisp-users/blob/a23e05fee407f739803740a0e800fc1fa3ad33f4/openwisp_users/tests/test_admin.py#L1391-L195

openwisp_users/multitenancy.py
```diff
@@ -8,7 +8,34 @@
...
+            return
+        fields = form.base_fields
+        if 'organization' in fields:
+            org_field = fields['organization']
+            org_field.initial = request.user.selected_org(org_field)
+            org_field.empty_label = self._get_empty_level(
+                request.user.is_superuser, org_field
+            )
+
+    def _get_empty_level(self, superuser, org_field):
```
devkapilbansal, line 27: @niteshsinha17 it should be _get_empty_label

niteshsinha17: Oh! Thanks @devkapilbansal

niteshsinha17:
@nemesisdesign It is ready for review. 
Do I need to resolve this conflict by rebasing??

devkapilbansal:
> @nemesisdesign It is ready for review.
> Do I need to resolve this conflict by rebasing??

Yes @niteshsinha17 please solve the merge conflicts

niteshsinha17:
Re-based to current master. Ready for review.
I have one doubt. When org field is not required then we are setting ``empty label`` to ``Shared systemwide (no organization)``. but when there is only one organization and the org field is not required which should be chosen as initial value? Currently ``Shared systemwide (no organization)`` is shown as selected value because it is an empty label. @nemesisdesign Can you please answer this.

devkapilbansal:
@niteshsinha17 may be the only organization should be selected when there is only one organization. 🤔

niteshsinha17:
> @niteshsinha17 may be the only organization should be selected when there is only one organization. 🤔

Actually, there is only one org but it looks like has 2 org because of the empty label and empty label seems to be selected indeed it is not been selected. 
I think it is quite complicated. We should just ignore it for this case otherwise we will need to change lots of things.

openwisp_users/admin.py
```diff
@@ -106,13 +106,15 @@ def get_formset(self, request, obj=None, **kwargs):
...
+            self.select_organization(request, formset.form)
             return formset
         if not request.user.is_superuser:
             formset.form.base_fields[
                 'organization'
             ].queryset = Organization.objects.filter(
                 pk__in=request.user.organizations_managed
             )
+        self.select_organization(request, formset.form)
         return formset
```
devkapilbansal, line 118: Can we write it like this?
```suggestion
        if not request.user.is_superuser:
            formset.form.base_fields[
                'organization'
            ].queryset = Organization.objects.filter(
                pk__in=request.user.organizations_managed
            )
        self.select_organization(request, formset.form)
        return formset
```

openwisp_users/tests/test_admin.py
```diff
@@ -1658,3 +1658,24 @@ def test_organization_user_filter(self):
...
                 self.assertNotContains(response, f'<a href="?organization={pk}"')
+
+    def test_select_organization(self):
+        # test for superuser
+        self._login()
+        url = reverse(f'admin:{self.app_label}_user_add')
+        html = 'selected>default</option>'
+        response = self.client.get(url)
+        self.assertContains(response, html)
+        self._logout()
```
devkapilbansal, line 1669: Would be better if you use `subTest` here

devkapilbansal (changes requested):
@niteshsinha17 I have minor suggestions below :point_down:

openwisp_users/base/models.py
```diff
@@ -97,6 +97,24 @@ def is_owner_of_any_organization(self):
...
+                return None
+            return selected_org
+        org_queryset = org_field._queryset
+        if org_queryset.count() == 1:
+            selected_org = org_queryset[0]
+        else:
+            selected_org = False  # Tells selected org is None
```
atb00ker, line 114: I don't get it, here we are returning `False`, above we are doing 
```
if selected_org is False:
 return None
```

do we want to return None? Can you please explain this? :smile:

openwisp_users/multitenancy.py
```diff
@@ -67,15 +94,12 @@ def _edit_form(self, request, form):
         fields = form.base_fields
         user = request.user
         org_field = fields.get('organization')
-        if user.is_superuser and org_field and not org_field.required:
-            org_field.empty_label = _('Shared systemwide (no organization)')
```
atb00ker, line 71: Do we really want to remove this feature, would we need to give backward compatibility for it?

atb00ker (commented):
Great going, I haven't tested but some comments

openwisp_users/apps.py
```diff
@@ -83,6 +93,15 @@ def connect_receivers(self):
                 dispatch_uid='update_user_permissions',
             )
 
+    def update_selected_org(cls, user=None, **kwargs):
+        if user:
+            cache_key = 'selected_{}_org'.format(user.pk)
```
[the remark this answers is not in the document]
niteshsinha17: Thanks @atb00ker I was just trying to maintain consistence. I will change it.

openwisp_users/base/models.py
niteshsinha17: Thanks, I think it has been missed.

openwisp_users/multitenancy.py
niteshsinha17: I have not removed this. It is being used in `_get_empty_level` of class `SelectOrgMixin`

niteshsinha17:
@atb00ker I have made all the changes. Please check

openwisp_users/apps.py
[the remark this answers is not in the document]
atb00ker: Slowly but surely we will have to move those old strings as well! :smile:

nemesisdesign:
Superseded by https://github.com/openwisp/openwisp-users/issues/333, thanks to all who contributed to this effort :pray: .

PR #208, opened 2020-12-16
[feature] Allow authenticating with email and phone number #206

### LINKED ISSUE #206: [feature] Allow logging in with phone number, email or username
Right now is only possible to log in with the username, but nowadays email and phone numbers are widely used.

We need to create an authentication backend that allows authenticating with any of the following:

- username
- email
- phone number

And of course, password.

Important points:

- [ ] the string supplied for username, email or phone number must be not empty.
- [ ] we could use an OR query, in order of precedence: get user with this email or this phone number or this username
- [ ] it's possible the query will return multiple results, the precedence of the previous point should be respected: write a test in which user A has username: user@test.com and user B has email user@test.com, test the authentication backend and ensure user B gets authenticated correctly
- [ ] write a similar test like the one before but with a phone number

NoumbissiValere:
Closes #206

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -491,6 +491,20 @@ to avoid generating database queries each time is called.
     >>> user.has_permission('openwisp_users.add_user')
     ... True
 
+Authentication Backend
+----------------------
+The custom authentication backend: ``UserAuthenticationBackend`` allow a user to authenticate
+using an ``email`` or ``phone_number`` or ``username`` and a ``password`` as credentials.
+It can be used as follows:
+
+.. code-block:: python
+
+    from openwisp_users.backends import UserAuthenticationBackend
+
+    backend = UserAuthenticationBackend()
+    backend.authenticate(request, phone_number, password)
+
+
 Django REST Framework Permission Classes
 ----------------------------------------
 
--- a/README.rst
+++ b/README.rst
@@ -97,10 +97,14 @@ Setup (integrate in an existing django project)
         'drf_yasg',
     ]
 
-also add ``AUTH_USER_MODEL`` and ``SITE_ID`` to your ``settings.py``::
+also add ``AUTH_USER_MODEL``, ``SITE_ID`` and ``AUTHENTICATION_BACKENDS``
+to your ``settings.py``::
 
     AUTH_USER_MODEL = 'openwisp_users.User'
     SITE_ID = 1
+    AUTHENTICATION_BACKENDS = [
+        'openwisp_users.backends.UsersAuthenticationBackend',
+    ]
 
 ``urls.py``:
 
@@ -493,17 +497,21 @@ to avoid generating database queries each time is called.
 
 Authentication Backend
 ----------------------
-The custom authentication backend: ``UserAuthenticationBackend`` allow a user to authenticate
-using an ``email`` or ``phone_number`` or ``username`` and a ``password`` as credentials.
+
+The authentication backend in ``openwisp_users.backends.UsersAuthenticationBackend``
+allows users to authenticate using their
+``email`` or ``phone_number`` instead of their ``username``.
+Authenticating with the ``username`` is still allowed,
+but ``email`` and ``phone_number`` have precedence.
+
 It can be used as follows:
 
 .. code-block:: python
 
-    from openwisp_users.backends import UserAuthenticationBackend
+    from openwisp_users.backends import UsersAuthenticationBackend
 
-    backend = UserAuthenticationBackend()
+    backend = UsersAuthenticationBackend()
     backend.authenticate(request, phone_number, password)
-
 
 Django REST Framework Permission Classes
 ----------------------------------------
--- a/openwisp_users/backends.py
+++ b/openwisp_users/backends.py
@@ -5,25 +5,20 @@ from django.db.models import Q
 User = get_user_model()
 
 
-class UserAuthenticationBackend(ModelBackend):
+class UsersAuthenticationBackend(ModelBackend):
     def authenticate(self, request, username=None, password=None, **kwargs):
         queryset = self.get_users(username)
         try:
-            # can not use queryset.first() because it orders the queryset before
-            # returning the first object which is not what we want
+            # can not use queryset.first() because it orders the queryset
+            # by pk before returning the first object which is not what we want
             user = queryset[0]
         except IndexError:
             return None
-        else:
-            if user.check_password(password) and self.user_can_authenticate(user):
-                return user
+        if user.check_password(password) and self.user_can_authenticate(user):
+            return user
         return None
 
     def get_users(self, identifier):
-        """
-        returns users with the identifier supplied
-        """
-        users = User.objects.filter(
+        return User.objects.filter(
             Q(email=identifier) | Q(phone_number=identifier) | Q(username=identifier)
         )
-        return users
--- a/openwisp_users/backends.py
+++ b/openwisp_users/backends.py
@@ -1,5 +1,6 @@
 from django.contrib.auth import get_user_model
 from django.contrib.auth.backends import ModelBackend
+from django.core.exceptions import ValidationError
 from django.db.models import Q
 
 User = get_user_model()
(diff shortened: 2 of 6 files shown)


tests/openwisp2/settings.py
```diff
@@ -90,6 +90,11 @@
     }
 ]
 
+AUTHENTICATION_BACKENDS = [
+    'django.contrib.auth.backends.ModelBackend',
+    'openwisp_users.backends.UserAuthenticationBackend',
```
nemesisdesign, line 95: I think `openwisp_users.backends.UserAuthenticationBackend` should be enough

openwisp_users/backends.py
```diff
@@ -0,0 +1,16 @@
...
+
+User = get_user_model()
+
+
+class UserAuthenticationBackend(ModelBackend):
+    def authenticate(self, request, username=None, password=None, **kwargs):
+        users = User.objects.filter(
+            Q(email=username) | Q(phone_number=username) | Q(username=username)
+        )
+        for user in users:
```
nemesisdesign, line 13: this looks like a potential security issue, I think we should just get the first result of the queryset

nemesisdesign (changes requested):

openwisp_users/backends.py
NoumbissiValere: @nemesisdesign Please which security issue are talking about?
I don't think getting the first result of the queryset will be a good idea because, if we consider the case below:

```
user1 = self._create_user(
            username='tester',
            email='tester@gmail.com',
            phone_number='+237675579231',
            password='tester',
        )
        user2 = self._create_user(
            username='tester@gmail.com',
            email='tester1@gmail.com',
            phone_number='+237675579232',
            password='tester1',
        )
```

with the solution you proposed, 
`user2` will never be able to login if he provides `username and password` as credentials. this is because `user1` will be the first in the queryset since we are checking for email first. 

I believe user `passwords` are secret, that is the reason why in my solution, I marked a user as authenticated only if the password supplied matches with the user's password and the user is active. 

Another solution I thought about is adding a clean to `User model` which prevent users from signing up with a `username` which is either the `phone_number` or `email` of already existing user.

If these solutions are not good enough, Please explain to me the security issues you saw and why you think getting the first result is the best thing to do

openwisp_users/backends.py
```diff
@@ -0,0 +1,16 @@
...
+from django.db.models import Q
+
+User = get_user_model()
+
+
+class UserAuthenticationBackend(ModelBackend):
+    def authenticate(self, request, username=None, password=None, **kwargs):
+        users = User.objects.filter(
+            Q(email=username) | Q(phone_number=username) | Q(username=username)
+        )
```
nemesisdesign, line 12: Following up from https://github.com/openwisp/openwisp-radius/pull/207#discussion_r547051623

you could move these line into a method:

```python
def get_user(self, identifier):
    return User.objects.filter(
        Q(email=identifier) | Q(phone_number=identifier) | Q(username=identifier)
    )
```

Then in openwisp-radius you would call `auth_backend.get_user(username).first()`

openwisp_users/backends.py
nemesisdesign: `user2` is trying to pose as user1, such a user is not supposed to exist so it's not a problem at all if it can't log in.
This would also simplify the code which would not need a loop.

openwisp_users/backends.py
```diff
@@ -0,0 +1,35 @@
...
+        # can not use users.first() because it orders the queryset before
+        # returning the first object which is not what we want
+        user = users[0] if users else None
```
nemesisdesign, line 16: why aren't you calling `get_user` here?
We can't have two methods doing almost the same thing but differently and one method is not even used in openwisp-users.

What I asked is to get the common logic to filter the user into the `get_user` method, which would be called here and also in openwisp-radius (while openwisp-radius will need to add some more filtering if needed).

openwisp_users/backends.py
```diff
@@ -0,0 +1,35 @@
...
+        return None
+
+    def get_user(self, identifier):
+        """
+        returns an active user with the identifier supplied
+        """
+        try:
+            # By default, primary key is provided as identifier
+            # during login and logout
+            return User.objects.get(pk=identifier, is_active=True)
```
nemesisdesign, line 28: I really don't understand where this is coming from.

openwisp_users/backends.py
```diff
@@ -0,0 +1,35 @@
...
+        except (User.DoesNotExist, ValidationError):
+            users = User.objects.filter(
+                Q(email=identifier)
+                | Q(phone_number=identifier)
+                | Q(username=identifier)
+            ).filter(is_active=True)
+            return users[0] if users else None
```
nemesisdesign, line 35: I think it's preferable to directly `return users[0]`, which will use a LIMIT query (which is more efficient), wile catching the exception in case there's no user (which should be the rare case).

openwisp_users/backends.py
```diff
@@ -0,0 +1,35 @@
...
+        """
+        try:
+            # By default, primary key is provided as identifier
+            # during login and logout
+            return User.objects.get(pk=identifier, is_active=True)
+        except (User.DoesNotExist, ValidationError):
+            users = User.objects.filter(
+                Q(email=identifier)
+                | Q(phone_number=identifier)
+                | Q(username=identifier)
```
nemesisdesign, line 33: this logic is duplicated with the method above, please avoid it

openwisp_users/backends.py
```diff
@@ -0,0 +1,35 @@
...
+            ).filter(is_active=True)
```
nemesisdesign, line 34: are you adding this for openwisp-radius? If yes, it's wrong, you should instead chain the queryset on openwisp-radius, eg: `auth_backend.get_user().filter(is_active=True)`.

nemesisdesign (changes requested):
There's still work to do here, see below and please add a couple of lines in the README about this authentication backend, explaining that it is used to allow logging in with email, phone number or username (in order of precedence), please add it before the [DRF classes section](https://github.com/openwisp/openwisp-users#django-rest-framework-permission-classes).

openwisp_users/backends.py
NoumbissiValere: @nemesisdesign that [django doc](https://docs.djangoproject.com/en/3.1/ref/models/querysets/#first) explains what i said in the comment above. `.first()` orders the queryset before returning the first value. `If the QuerySet has no ordering defined, then the queryset is automatically ordered by the primary key`. So using `.first()` is a problem because the first element might not be element we wanted in this case. I experienced such behaviors during testing.

openwisp_users/backends.py
NoumbissiValere: This is here because, during login and logout, django calls the `get_user` method and pass to it the `pk` of the user. infact the built in `get_user` inherited from the parent backend accepts a `pk` which it uses to filter for the user. I think I will create a new method rather than override the `get_user` to avoid this kind of confusion now and also in the future

README.rst
```diff
@@ -97,10 +97,14 @@ Setup (integrate in an existing django project)
...
 
-also add ``AUTH_USER_MODEL`` and ``SITE_ID`` to your ``settings.py``::
+also add ``AUTH_USER_MODEL``, ``SITE_ID`` and ``AUTHENTICATION_BACKENDS``
+to your ``settings.py``::
 
     AUTH_USER_MODEL = 'openwisp_users.User'
     SITE_ID = 1
+    AUTHENTICATION_BACKENDS = [
+        'openwisp_users.backends.UsersAuthenticationBackend',
+    ]
```
nemesisdesign, line 107: @NoumbissiValere please don't forget to add information like this one to the README: every time we add something to the test env we have to add it to the setup instructions too.

openwisp_users/backends.py
```diff
@@ -5,25 +5,20 @@
...
-            # can not use queryset.first() because it orders the queryset before
-            # returning the first object which is not what we want
+            # can not use queryset.first() because it orders the queryset
+            # by pk before returning the first object which is not what we want
             user = queryset[0]
         except IndexError:
             return None
-        else:
-            if user.check_password(password) and self.user_can_authenticate(user):
-                return user
```
nemesisdesign: avoid unnecessary nesting

nemesisdesign (approved):

PR #209, opened 2020-12-28
[admin] Change empty org label #171

### LINKED ISSUE #171: [admin] Add a human readable value for empty org selection
When the organization selection is optional, it means that leaving it empty makes that object available for all the organizations in the system.

Right now, this is not intuitive at all. We could use some admin form magic to replace the empty string with something like: `Shared systemwide (no organization)`.

Purhan:
Closes #171

### DIFF
--- a/openwisp_users/multitenancy.py
+++ b/openwisp_users/multitenancy.py
@@ -76,6 +76,7 @@ class MultitenantAdminMixin(object):
             if 'organization' in fields:
                 org_field = fields['organization']
                 org_field.queryset = org_field.queryset.filter(pk__in=orgs_pk)
+                org_field.empty_label = None
             # other relations
             q = Q(organization__in=orgs_pk) | Q(organization=None)
             for field_name in self.multitenant_shared_relations:
--- a/openwisp_users/multitenancy.py
+++ b/openwisp_users/multitenancy.py
@@ -65,16 +65,15 @@ class MultitenantAdminMixin(object):
         else show everything
         """
         fields = form.base_fields
-        if 'organization' in fields and not fields['organization'].required:
-            fields['organization'].empty_label = _(
-                'Shared systemwide (no organization)'
-            )
-        if not request.user.is_superuser:
-            orgs_pk = request.user.organizations_managed
+        user = request.user
+        org_field = fields.get('organization')
+        if user.is_superuser and org_field and not org_field.required:
+            org_field.empty_label = _('Shared systemwide (no organization)')
+        elif not user.is_superuser:
+            orgs_pk = user.organizations_managed
             # organizations relation;
             # may be readonly and not present in field list
-            if 'organization' in fields:
-                org_field = fields['organization']
+            if org_field:
                 org_field.queryset = org_field.queryset.filter(pk__in=orgs_pk)
                 org_field.empty_label = None
             # other relations
@@ -103,8 +102,8 @@ class MultitenantAdminMixin(object):
         from same organization and hide superusers
         if superuser is logged in - show all users
         """
-        if not request.user.is_superuser:
-            user = request.user
+        user = request.user
+        if not user.is_superuser:
             org_users = OrganizationUser.objects.filter(user=user).select_related(
                 'organization'
             )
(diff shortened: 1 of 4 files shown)


Purhan:
image
This is how it looks right now.

In my opinion, this should be different when organization field is required. Otherwise it might leave the impression that systemwide is available but then it'd raise validation error while submitting the form. I was thinking we could do this:
```python
if fields['organization'].required:
    # Do something here to change the label, like a strikethrough
    fields['organization'].empty_label = '\u0336'.join(
        'Shared systemwide (no organization)'
    )
```
To add a ~~strike~~ to indicate that this option is not available.

openwisp_users/multitenancy.py
```diff
@@ -65,6 +65,8 @@ def _edit_form(self, request, form):
         else show everything
         """
         fields = form.base_fields
+        if 'organization' in fields:
+            fields['organization'].empty_label = 'Shared systemwide (no organization)'
```
pandafy, line 69: Make this string translatable like following:

https://github.com/openwisp/openwisp-users/blob/84e7cd145f5243b452c94f14d62ae92e93557483/openwisp_users/multitenancy.py#L35

pandafy (changes requested):
> In my opinion, this should be different when organization field is required. Otherwise it might leave the impression that systemwide is available but then it'd raise validation error while submitting the form.

I totally agree with you @Purhan, but the code you have suggested can be modified as below. 

```python
if not fields['organization'].required:
    fields['organization'].empty_label = _('Shared systemwide (no organization)')
```

I don't think, we need to add strikethrough to show that a field is required. We can keep the default `empty_label` when the organization field is required.

openwisp_users/multitenancy.py
Purhan: @pandafy sure. One more thing, as you can see I wrote a test for this, but after implementing these changes, I can't test this feature as there is no model with optional organization field in this project.

Purhan:
@pandafy Done. Like I said, I removed the test as well.

pandafy (changes requested):
> @pandafy Done. Like I said, I removed the test as well.

@Purhan tests are good. Let's add a test.

> @pandafy sure. One more thing, as you can see I wrote a test for this, but after implementing these changes, I can't test this feature as there is no model with optional organization field in this project.

I was not able to reply to this before. But better later than never. 

You can add an admin site for `Template` model below. Since it uses `ShareableOrgMixin`, the organization field is optional.

https://github.com/openwisp/openwisp-users/blob/4569b605d2402eed75bcf15e100dad9e01462c67/tests/testapp/models.py#L9-L13


@devkapilbansal the optional organization field is not used directly in this module, but in other modules of OpenWISP. Sometimes, you will need to think about OpenWISP as whole instead of individual module. Then only, some features make sense.

Purhan:
> You can add an admin site for Template model below. 

`Template` is also not registered on the admin site by default. Should I register it?

openwisp_users/multitenancy.py
```diff
@@ -65,6 +65,10 @@ def _edit_form(self, request, form):
         else show everything
         """
         fields = form.base_fields
+        if 'organization' in fields and not fields['organization'].required:
+            fields['organization'].empty_label = _(
+                'Shared systemwide (no organization)'
+            )
```
nemesisdesign, line 71: this should be done only if the user is a superuser (should be the first check)

openwisp_users/multitenancy.py
```diff
@@ -65,6 +65,10 @@ def _edit_form(self, request, form):
...
         if not request.user.is_superuser:
             orgs_pk = request.user.organizations_managed
```
nemesisdesign, line 73: here we'll need to add `fields['organization'].empty_label = None`

nemesisdesign (changes requested):
Great work @Purhan, looks good but while testing it I realized there's a few things to improve which are not related to this PR.

The empty label should be available only to super users (I found out is simple: `fields['organization'].empty_label = None`, let me know if you can help with this (we'll need a test for this).

nemesisdesign (approved):
Great work @Purhan :+1: :+1:

PR #211, opened 2021-01-05
[fix] Fixed test_organization_default_label breaking integration tests

nemesisdesign:
The test "test_organization_default_label" depends on a test app
which is not part of openwisp-users, therefore this test should
be shipped in the test app rather than openwisp-users, because
the tests of openwisp-users are imported and re-run in other
modules as integration tests, which test_organization_default_label
was breaking.

@Purhan see :point_up:  and https://github.com/openwisp/openwisp-radius/runs/1628664181.


Purhan:
I see, I will keep this in mind for the future.

Purhan:
From discussion on IM, `test_login_page` is also failing for likely the same reason.

PR #215, opened 2021-01-10
[fix] Fix reversion registration for Group #214

### LINKED ISSUE #214: [bug] Group has not been registered with django-reversion
This shows up in instances installed via ansible-openwisp2 but can be replicated in the dev env.

**How to replicate**:

- add `reversion` to INSTALLED_APPS, install django-reversion if you don't have it installed in your virtualenv already, in openwisp-users is optional
- go to groups, open a group, (eg: administrator), make some changes, save, repeat the process again
- click on the "history" button in the upper right part of the page
- different revisions will appear, click on any

**Expected outcome**:

Page with the revision appears.

**Actual outcome**:

```
<class 'django.contrib.auth.models.Group'> has not been registered with django-reversion
```

Purhan:
Closes #214


openwisp_users/admin.py
```diff
@@ -470,7 +470,9 @@ def save_formset(self, request, form, formset, change):
 
 
 class GroupAdmin(BaseGroupAdmin, BaseAdmin):
-    pass
+    def reversion_register(self, model, **kwargs):
+        # Correctly register the proxy model
+        return super().reversion_register(model, for_concrete_model=False, **kwargs,)
```
Purhan, line 475: @nemesisdesign I had to include this patch here since we are using a proxy model for Groups. We could instead use an `AbstractGroup` class like we are using for other auth models like User and Organization, but that'd be backward incompatible.

Also, I'm not sure how to test this in this module.

openwisp_users/admin.py
```diff
@@ -470,7 +470,10 @@ def save_formset(self, request, form, formset, change):
 
 
 class GroupAdmin(BaseGroupAdmin, BaseAdmin):
-    pass
+    if 'reversion' in settings.INSTALLED_APPS:
+        # Correctly register the proxy model
+        def reversion_register(self, model, **kwargs):
+            return super().reversion_register(model, for_concrete_model=False, **kwargs)
```
Purhan, line 476: @nemesisdesign I had to include this patch here since we are using a proxy model for Groups. We could instead use an Abstract `BaseGroup` class (not to be confused with `AbstractGroup`) like we are using for other auth models like User and Organization, but that'd be backward incompatible.

Also, I'm not sure how to test this in this module.

nemesisdesign: Let's test this by adding django-reversion to requirements-test.txt and to INSTALLED_APPS in the test project, we need to make sure we won't get regressions in the future. What do you think?

nemesisdesign (changes requested):
Thanks @Purhan! See below.

tests/testapp/tests/test_admin.py
```diff
@@ -36,7 +38,22 @@ def test_organization_default_label(self):
             self.assertNotContains(
                 r, 'Shared systemwide (no organization)',
             )
-
+            
+    def test_group_reversion(self):
```
Purhan, line 42: @nemesisdesign I had this idea:
```python
with self.modify_settings(
    INSTALLED_APPS={
        'append': 'reversion'
    }
):
```
unfortunately, we cannot directly add this in tests because URLs from `reversion` will not be loaded at runtime, which this test depends on. So I added it directly in `settings.py` with a condition to remove when tests are not running (4fa8427e5143376f7533a7f18a09543ddaf2855b).
This had one side effect, it increased the number of queries for some tests (I updated that in aa7f3e197af768891be9875d4d787543102d347a)

nemesisdesign: That's good thanks, you can add it to the PR.

openwisp_users/tests/test_admin.py
```diff
@@ -1353,7 +1353,7 @@ def test_delete_org_user(self):
             )
             self.assertContains(r, msg)
             post_data.update({'post': 'yes'})
-            with self.assertNumQueries(17):
+            with self.assertNumQueries(21):
```
nemesisdesign, line 1356: can you add a comment before each assertNumQueries like:

`# django-reversion adds ~4 queries`

Purhan: Done :+1:

openwisp_users/admin.py
Purhan: I believe we shouldn't see any regressions. From the [docs](https://django-reversion.readthedocs.io/en/stable/api.html#registration-api):
```
for_concrete_model=True
    If True proxy models will be saved under the same content type as their concrete model. If False, proxy models will be saved under their own content type, effectively giving proxy models their own distinct history.
```

nemesisdesign (commented):
Looks good, will test asap :+1:

nemesisdesign (approved):
Great work @Purhan :+1:

PR #217, opened 2021-01-11
[feature] Add API filter mixins #210

### LINKED ISSUE #210: [feature] Add API classes to filter by organization and to check parent objects
We need DRF API classes/mixins to perform the following operations:

- organization access:
  - filter objects related to the organizations the user has access to
  - filter objects related to the organizations the user can manage
  - filter objects related to the organizations the user owns
- parent object check
  - same as above but while filtering the parent object

The concept has been explored and implemented in https://github.com/openwisp/openwisp-ipam/blob/281485cad98bb1d961662f5628a370e9c7f16907/openwisp_ipam/api/utils.py#L5-L51.

We need to move that logic here, add a few more mixins (only the "managed" mixins are implemented in there) and add basic information to the README (eg: a new section called "Django REST Framework mixins" after the [Django REST Framework permission classes](https://github.com/openwisp/openwisp-ipam/blob/281485cad98bb1d961662f5628a370e9c7f16907/openwisp_ipam/api/utils.py#L5-L51)).

**Addendum**: we should also introduce reusable classes to make the browsable API relations multi-tenant aware.

Purhan:
Closes #210

### DIFF
--- a/openwisp_users/api/filters.py
+++ b/openwisp_users/api/filters.py
@@ -1,0 +1,183 @@
+import swapper
+from django.contrib.auth.models import Permission
+from django.core.exceptions import ValidationError
+from rest_framework.exceptions import NotFound, PermissionDenied
+
+Organization = swapper.load_model('openwisp_users', 'Organization')
+
+
+class FilterByOrganization:
+    """
+    Filter queryset based on the access to the organization
+    of the associated model. Use on of the sub-classes
+    """
+
+    def get_queryset(self):
+        qs = super().get_queryset()
+        if self.request.user.is_superuser:
+            return qs
+        return self.get_organization_queryset(qs)
+
+    def get_organization_queryset(self):
+        raise NotImplementedError()
+
+
+class FilterByOrganizationMembership(FilterByOrganization):
+    """
+    Filter queryset by organizations the user is a member of
+    """
+
+    def get_organization_queryset(self, qs):
+        return qs.filter(organization__in=self.request.user.organizations_dict)
+
+
+class FilterByOrganizationManaged(FilterByOrganization):
+    """
+    Filter queryset by organizations managed by user
+    """
+
+    def get_organization_queryset(self, qs):
+        return qs.filter(organization__in=self.request.user.organizations_managed)
+
+
+class FilterByOrganizationOwned(FilterByOrganization):
+    """
+    Filter queryset by organizations owned by user
+    """
+
+    def get_organization_queryset(self, qs):
+        return qs.filter(organization__in=self.request.user.organizations_owned)
+
+
+class FilterByParent:
+    """
+    Filter queryset based on one of the parent objects
+    """
+
+    def get_queryset(self):
+        qs = super().get_queryset()
+        self.assert_parent_exists()
+        return qs
+
+    def assert_parent_exists(self):
+        parent_queryset = self.get_parent_queryset()
+        if not self.request.user.is_superuser:
+            parent_queryset = self.get_organization_queryset(parent_queryset)
+        try:
+            assert parent_queryset.exists()
+        except (AssertionError, ValidationError):
+            raise NotFound(detail='No relevant data found.')
+
+    def get_parent_queryset(self):
+        raise NotImplementedError()
+
+    def get_organization_queryset(self):
+        raise NotImplementedError()
+
+
+class FilterByParentMembership(FilterByParent):
+    """
+    Filter queryset based on parent organization membership
+    """
+
+    def get_organization_queryset(self, qs):
+        return qs.filter(organization__in=self.request.user.organizations_dict)
+
+
+class FilterByParentManaged(FilterByParent):
+    """
+    Filter queryset based on parent organizations managed by user
+    """
+
+    def get_organization_queryset(self, qs):
+        return qs.filter(organization__in=self.request.user.organizations_managed)
+
+
+class FilterByParentOwned(FilterByParent):
+    """
+    Filter queryset based on parent organizations owned by user
+    """
+
+    def get_organization_queryset(self, qs):
+        return qs.filter(organization__in=self.request.user.organizations_owned)
+
+
+class FilterSerializerByOrganization:
+    """
+    Filter the options in browsable API for serializers
+    """
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        if self.context['request'].user.is_superuser:
+            return
+        self.filter_fields()
+
+    def filter_fields(self):
+        raise NotImplementedError()
+
+
+class FilterSerializerByOrgMembership(FilterSerializerByOrganization):
+    """
+    Filter serializer by organizations the user is member of
+    """
+
+    def filter_fields(self):
+        user = self.context['request'].user
+        organization_filter = user.organizations_dict
+        for field in self.fields:
+            if field == 'organization':
+                self.fields[field].queryset = self.fields[field].queryset.filter(
+                    pk__in=organization_filter
+                )
+                continue
+            try:
+                self.fields[field].queryset = self.fields[field].queryset.filter(
+                    organization__in=organization_filter
+                )
+            except AttributeError:
+                pass
+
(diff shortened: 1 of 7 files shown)


Purhan:
WIP: Have to add tests and probably some improvements in the code itself.

nemesisdesign (commented):
Great progress @Purhan! :+1:

See my comments below.

openwisp_users/api/filters.py
```diff
@@ -0,0 +1,157 @@
...
+        return qs.filter(organization__in=self.request.user.organizations_managed)
+
+
+class FilterByParentOwned(FilterByParent):
+    """
+    Filter queryset based on parent organizations owned by user
+    """
+
+    def get_organization_queryset(self, qs):
+        return qs.filter(organization__in=self.request.user.organizations_owned)
```
nemesisdesign, line 101: same concept here, you can eliminate the repetition of these 3 classes as you did below

nemesisdesign (changes requested):
Good progress @Purhan, I left another comment below.

Regarding tests, I suggest using the [testapp](https://github.com/openwisp/openwisp-users/tree/master/tests/testapp).

Purhan:
Alright, I'm back :smile:. @nemesisdesign I worked on this today, please take a look.

openwisp_users/api/filters.py
```diff
@@ -0,0 +1,159 @@
...
+        organization_filter = getattr(user, self._user_attr)
+        for field in self.fields:
+            if field == 'organization':
+                self.fields[field].queryset = self.fields[field].queryset.filter(
+                    pk__in=organization_filter
+                )
+                continue
+            try:
+                self.fields[field].queryset = self.fields[field].queryset.filter(
+                    organization__in=organization_filter
```
nemesisdesign, line 128: We may need the `organization_lookup` attribute here too, what do you think?

openwisp_users/api/filters.py
```diff
@@ -0,0 +1,159 @@
...
+        parent_queryset = self.get_parent_queryset()
+        if not self.request.user.is_superuser:
+            parent_queryset = self.get_organization_queryset(parent_queryset)
+        try:
+            assert parent_queryset.exists()
+        except (AssertionError, ValidationError):
+            raise NotFound(detail='No relevant data found.')
+
+    def get_organization_queryset(self, qs):
+        return qs.filter(organization__in=getattr(self.request.user, self._user_attr))
```
nemesisdesign, line 76: I don't think we'll need `organization_lookup` here because usually parent objects have the `organization` field defined on them, so this looks good :+1:

openwisp_users/api/filters.py
```diff
@@ -0,0 +1,159 @@
...
+        return qs
+
+    def assert_parent_exists(self):
+        parent_queryset = self.get_parent_queryset()
+        if not self.request.user.is_superuser:
+            parent_queryset = self.get_organization_queryset(parent_queryset)
+        try:
+            assert parent_queryset.exists()
+        except (AssertionError, ValidationError):
+            raise NotFound(detail='No relevant data found.')
```
nemesisdesign, line 73: can you change this to just `Not found.` please?

openwisp_users/api/filters.py
```diff
@@ -0,0 +1,159 @@
...
+        raise NotImplementedError()
+
+    def get_queryset(self):
+        qs = super().get_queryset()
+        if self.request.user.is_superuser:
+            return qs
+        return self.get_organization_queryset(qs)
+
+    def get_organization_queryset(self, qs):
+        return qs.filter(organization__in=getattr(self.request.user, self._user_attr))
```
nemesisdesign, line 25: There's a little issue here, some models do not have the `organization` relation on their model but on a parent.
In those cases this won't work.

For that reason, I suggest adding another class attribute called `organization_lookup`, which defaults to `organization__in`.

```python
class FilterByOrganization:
    # ...

    def get_organization_queryset(self, qs):
        return qs.filter(**{
            self.organization_lookup: getattr(self.request.user, self._user_attr)
        })
```

openwisp_users/api/filters.py
```diff
@@ -0,0 +1,159 @@
+import swapper
```
nemesisdesign, line 1: Let's not call this file `filters`, because that would be confused with other type of classes which are called filter and are a different thing (https://www.django-rest-framework.org/api-guide/filtering/#generic-filtering).

I suggest naming this file `mixins.py`, remember to update the docs.

tests/testapp/urls.py
```diff
@@ -9,4 +9,34 @@
     path('base_org_view', views.base_org_view, name='test_base_org_permission_view'),
     path('org_field_view', views.org_field_view, name='test_organization_field_view'),
     path('error_field_view', views.error_field_view, name='test_error_field_view'),
+    path(
+        'books_list_member_view/<str:shelf_id>',
```
nemesisdesign, line 13: It's a shame the URLs in this file are not realistic, let's not follow that convention.

Let's try to use realistic, RESTful URLs which will help the readers of the code to understand what these mixins are for.

This is a view which has a parent right? So the URL should be something like:

`/member/shelf/<str:shelf_id>/books/`

tests/testapp/urls.py
```diff
@@ -9,4 +9,34 @@
...
+        views.books_list_member_view,
+        name='test_books_list_member_view',
+    ),
+    path(
+        'books_list_manager_view/<str:shelf_id>',
```
nemesisdesign, line 18: `/manager/shelf/<str:shelf_id>/books/`

tests/testapp/urls.py
```diff
@@ -9,4 +9,34 @@
...
+        views.books_list_manager_view,
+        name='test_books_list_manager_view',
+    ),
+    path(
+        'books_list_owner_view/<str:shelf_id>',
```
nemesisdesign, line 23: `/owner/shelf/<str:shelf_id>/books/`

tests/testapp/urls.py
```diff
@@ -9,4 +9,34 @@
...
+        views.books_list_owner_view,
+        name='test_books_list_owner_view',
+    ),
+    path(
+        'shelf_list_member_view',
```
nemesisdesign, line 28: `/member/shelf/`

tests/testapp/urls.py
```diff
@@ -9,4 +9,34 @@
...
+        views.shelf_list_member_view,
+        name='test_shelf_list_member_view',
+    ),
+    path(
+        'shelf_list_manager_view',
```
nemesisdesign, line 33: `/manager/shelf/`

tests/testapp/urls.py
```diff
@@ -9,4 +9,34 @@
...
+        views.shelf_list_manager_view,
+        name='test_shelf_list_manager_view',
+    ),
+    path(
+        'shelf_list_owner_view',
```
nemesisdesign, line 38: `/owner/shelf/`

nemesisdesign (changes requested):
Thank you @Purhan!

I took some time to test it and review properly, since this is an important change.

The documentation was not super clear to me, so I tried to improve it.

The mixin to check/filter the parent was not explained properly, I am afraid you may have not understood fully why we are adding it, so I tried to make it clearer and I also left some comments in the test app regarding this aspect, let me know if it's clear and if not don't hesitate to ask questions.

Purhan:
@nemesisdesign Done, please take a look.

Purhan:
@nemesisdesign this is ready for review

nemesisdesign (changes requested):
Great work Purhan, can you rebase on the current master please?

Now is the time to go back to https://github.com/openwisp/openwisp-ipam/pull/76 and change its code to use this work, so we can validate it. Once that PR can work with this code without issues, we can merge both! :+1: :+1:

Purhan:
@nemesisdesign done rebasing. This is also ready for review now.

nemesisdesign (approved):
Great work @Purhan! :+1: :+1: :tada:

PR #223, opened 2021-01-13
[tests] Update TestMultitenantAdminMixin to use default operator grou…

### LINKED ISSUE #106: [tests] Update TestMultitenantAdminMixin to use default operator group
In this test utility code (which is reused across openwisp modules):

https://github.com/openwisp/openwisp-users/blob/cf5cd64f7d3c3515fcd0fc1f23077620b5a85964/openwisp_users/tests/utils.py#L23-L39

The test expects the permission of the operator to be defined as a class attribute.

This is obsolete, because in more recent versions of OpenWISP we added two default `auth.Group instances`: operator and administrator.

- update the test method to assign those groups
- remove `operator_permission_filters = []`
- remove `get_operator_permissions`

Open a pull request in each openwisp module to remove the `operator_permission_filters` and ensuring the new code works: https://github.com/search?q=org%3Aopenwisp+operator_permission_filters&type=Code

lehone-hp:
…p #106

- Update the _create_operator test method to assign default user groups
- Removed operator_permission_filters = []
- Remove get_operator_permissions

Fixes #106

### DIFF
--- a/openwisp_users/tests/utils.py
+++ b/openwisp_users/tests/utils.py
@@ -43,7 +43,13 @@ class TestMultitenantAdminMixin(object):
     def _logout(self):
         self.client.logout()
 
-    def _create_operator(self, organizations=[], **kwargs):
+    def _add_permissions(self, user, permission_filters):
+        filters = Q()
+        for filter in permission_filters:
+            filters = filters | Q(**filter)
+        user.user_permissions.add(*Permission.objects.filter(filters))
+
+    def _create_operator(self, organizations=[], permission_filters=[], **kwargs):
         opts = dict(
             username='operator',
             password='tester',
@@ -52,12 +58,9 @@ class TestMultitenantAdminMixin(object):
         )
         opts.update(kwargs)
         operator = User.objects.create_user(**opts)
-        groups = Group.objects.filter(name__in=['Administrator', 'Operator'])
+        groups = Group.objects.filter(name='Operator')
         operator.groups.set(groups)
-        permissions = Permission.objects.filter(
-            Q(codename__endswith='book') | Q(codename__endswith='shelf')
-        )
-        operator.user_permissions.add(*permissions)
+        self._add_permissions(operator, permission_filters)
         for organization in organizations:
             OrganizationUser.objects.create(
                 user=operator, organization=organization, is_admin=True
--- a/openwisp_users/tests/utils.py
+++ b/openwisp_users/tests/utils.py
@@ -9,6 +9,7 @@ from swapper import load_model
 Organization = load_model('openwisp_users', 'Organization')
 OrganizationOwner = load_model('openwisp_users', 'OrganizationOwner')
 OrganizationUser = load_model('openwisp_users', 'OrganizationUser')
+Group = load_model('openwisp_users', 'Group')
 User = get_user_model()
 
 
@@ -42,14 +43,6 @@ class TestMultitenantAdminMixin(object):
     def _logout(self):
         self.client.logout()
 
-    operator_permission_filters = []
-
-    def get_operator_permissions(self):
-        filters = Q()
-        for filter in self.operator_permission_filters:
-            filters = filters | Q(**filter)
-        return Permission.objects.filter(filters)
-
     def _create_operator(self, organizations=[], **kwargs):
         opts = dict(
             username='operator',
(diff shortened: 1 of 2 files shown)


openwisp_users/tests/utils.py
```diff
@@ -57,7 +50,13 @@ def _create_operator(self, organizations=[], **kwargs):
         )
         opts.update(kwargs)
         operator = User.objects.create_user(**opts)
-        operator.user_permissions.add(*self.get_operator_permissions())
+        groups = Group.objects.filter(name__in=['Administrator', 'Operator'])
+        operator.groups.set(groups)
+        permissions = Permission.objects.filter(
+            Q(codename__endswith='book') | Q(codename__endswith='shelf')
+        )
+        for permission in permissions:
```
NoumbissiValere, line 58: you remove this for loop and user `operator.user_permissions.add(*permission)`

NoumbissiValere (changes requested):
Good progress @lehone-hp

openwisp_users/tests/utils.py
lehone-hp: okay.

nemesisdesign (commented):
@lehone-hp this looks good but since this change will impact other modules, we need to update the other modules too (as mentioned in the issue description), the modules impacted by this change can be found here: https://github.com/search?q=org%3Aopenwisp+operator_permission_filters&type=Code

We could start with openwisp-controller:

You'll need to change the dependency on openwisp-users as we did here: https://github.com/openwisp/openwisp-radius/blob/ee5220f9aebc81508582fb1d6fa5f5338dd51c90/requirements.txt#L7-L9, with the difference that you need to point to your forked branch (`https://github.com/lehone-hp/openwisp-users/tarball/issue-106`).

lehone-hp:
> @lehone-hp this looks good but since this change will impact other modules, we need to update the other modules too (as mentioned in the issue description), the modules impacted by this change can be found here: https://github.com/search?q=org%3Aopenwisp+operator_permission_filters&type=Code
> 
> We could start with openwisp-controller:
> 
> You'll need to change the dependency on openwisp-users as we did here: https://github.com/openwisp/openwisp-radius/blob/ee5220f9aebc81508582fb1d6fa5f5338dd51c90/requirements.txt#L7-L9, with the difference that you need to point to your forked branch (`https://github.com/lehone-hp/openwisp-users/tarball/issue-106`).

Okya, working on it now.

lehone-hp:
@nemesisdesign I have opened three other PRs to complete the second part of this issue:
https://github.com/openwisp/openwisp-controller/pull/371
https://github.com/openwisp/openwisp-radius/pull/226
https://github.com/openwisp/openwisp-network-topology/pull/104

openwisp_users/tests/utils.py
```diff
@@ -57,7 +50,12 @@ def _create_operator(self, organizations=[], **kwargs):
         )
         opts.update(kwargs)
         operator = User.objects.create_user(**opts)
-        operator.user_permissions.add(*self.get_operator_permissions())
+        groups = Group.objects.filter(name__in=['Administrator', 'Operator'])
+        operator.groups.set(groups)
+        permissions = Permission.objects.filter(
+            Q(codename__endswith='book') | Q(codename__endswith='shelf')
+        )
```
atb00ker, line 57: I see, it seems we need to add additional permissions to the operator, but wouldn't it be better that 'shelf' and 'book' come from an optional class variable? 

Okay, I see that Fed has asked in the issue to remove `operator_permission_filter` but hardcoding it might not be the best thing because if a module needs to add more permissions or remove these permissions, this wouldn't be ideal. What do you think @nemesisdesign ?

openwisp_users/tests/utils.py
```diff
@@ -57,7 +50,12 @@ def _create_operator(self, organizations=[], **kwargs):
         )
         opts.update(kwargs)
         operator = User.objects.create_user(**opts)
-        operator.user_permissions.add(*self.get_operator_permissions())
+        groups = Group.objects.filter(name__in=['Administrator', 'Operator'])
```
atb00ker, line 53: I think I've not understood the issue or solution properly.
If we are creation an operator why `name__in=['Administrator', 'Operator']`, doesn't it just need to be in the Operator group?

atb00ker (commented):
Hi, great progress @lehone-hp ! :-)

openwisp_users/tests/utils.py
lehone-hp: The issue required that I should assign both Administrator and Operator groups.

openwisp_users/tests/utils.py
lehone-hp: The [issue](https://github.com/openwisp/openwisp-users/issues/106) specified that I should assign both Administrator and Operator groups.

codesankalp:
Resolved in https://github.com/openwisp/openwisp-users/pull/309

PR #230, opened 2021-02-06
[fix] Fix AttributeError for filter classes #229

### LINKED ISSUE #229: [bug] Don't allow filter classes to be used without authentication
https://github.com/openwisp/openwisp-controller/pull/374#pullrequestreview-581968990

If used in views without `IsAuthenticated` permission, the filter classes will raise an exception.

Purhan:
Fixes #229

### DIFF
--- a/openwisp_users/api/mixins.py
+++ b/openwisp_users/api/mixins.py
@@ -11,14 +11,12 @@ class FilterByOrganization:
     Filter queryset based on the access to the organization
     of the associated model. Use on of the sub-classes
     """
+    permission_classes = [IsAuthenticated]
+    organization_lookup = 'organization__in'
 
     @property
     def _user_attr(self):
         raise NotImplementedError()
-
-    organization_lookup = 'organization__in'
-
-    permission_classes = [IsAuthenticated]
 
     def get_queryset(self):
         qs = super().get_queryset()
@@ -60,12 +58,11 @@ class FilterByParent:
     """
     Filter queryset based on one of the parent objects
     """
+    permission_classes = [IsAuthenticated]
 
     @property
     def _user_attr(self):
         raise NotImplementedError()
-
-    permission_classes = [IsAuthenticated]
 
     def get_queryset(self):
         qs = super().get_queryset()
@@ -116,12 +113,11 @@ class FilterSerializerByOrganization:
     """
     Filter the options in browsable API for serializers
     """
+    organization_lookup = 'organization__in'
 
     @property
     def _user_attr(self):
         raise NotImplementedError()
-
-    organization_lookup = 'organization__in'
 
     def filter_fields(self):
         user = self.context['request'].user
--- a/openwisp_users/api/mixins.py
+++ b/openwisp_users/api/mixins.py
@@ -11,6 +11,7 @@ class FilterByOrganization:
     Filter queryset based on the access to the organization
     of the associated model. Use on of the sub-classes
     """
+
     permission_classes = [IsAuthenticated]
     organization_lookup = 'organization__in'
 
@@ -58,6 +59,7 @@ class FilterByParent:
     """
     Filter queryset based on one of the parent objects
     """
+
     permission_classes = [IsAuthenticated]
 
     @property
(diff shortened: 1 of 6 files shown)


tests/testapp/urls.py
```diff
@@ -37,4 +37,14 @@
     path(
         'owner/shelf', views.shelf_list_owner_view, name='test_shelf_list_owner_view',
     ),
+    path(
+        'owner/shelf/<str:shelf_id>/books',
```
nemesisdesign, line 41: there's a missing trailing slash here

openwisp_users/api/mixins.py
```diff
@@ -17,6 +18,9 @@ def _user_attr(self):
 
     organization_lookup = 'organization__in'
 
+    def get_permissions(self):
+        return super().get_permissions() + [IsAuthenticated()]
```
nemesisdesign, line 22: what about defining it in `permission_classes`? Eg:

```python
permission_classes = [IsAuthenticated]
```

This way we can override it if needed: we may need a custom `IsAuthenticated` or anything else, in that case we can override the behavior, while with this code overriding will be harder.

What do you think?

openwisp_users/api/mixins.py
```diff
@@ -62,6 +66,9 @@ class FilterByParent:
     def _user_attr(self):
         raise NotImplementedError()
 
+    def get_permissions(self):
+        return super().get_permissions() + [IsAuthenticated()]
```
nemesisdesign, line 70: same here

nemesisdesign (changes requested):
Great @Purhan, see my comments below.

2 questions:

- is it possible to add 1 test which fails without the `IsAuthenticated` permission? Like the failure we had in ow-controller
- can you please update the docs in the README to mention `IsAuthenticated` is shipped by default in these classes, since it requires the user to be authenticated to work.

tests/testapp/urls.py
Purhan: All other paths in this file don't have a trailing slash, should I just add it to all?

openwisp_users/api/mixins.py
Purhan: My idea was to make it robust so that the developers wouldn't have to worry about adding this permission class separately while creating views. But sure, we can do this instead.

For the time being, I'm keeping it simple as you said:
```python
permission_classes = [IsAuthenticated]
```

Also, I had this other idea, we could check if the user is authenticated, manually:
```python
...
if not (request.user and request.user.is_authenticated):
            raise NotFound()
...
```
Let me know if this will be better.

nemesisdesign (commented):
Looks good, will test asap.

Purhan:
Forgot to update the documentation, will do soon

tests/testapp/tests/test_filter_classes.py
```diff
@@ -177,3 +178,23 @@ def test_browsable_api_filter_member(self):
...
+        r = self.client.get(reverse("test_shelf_list_unauthorized_view"))
+        self.assertEqual(r.status_code, 401)
+        r = self.client.get(
+            reverse("test_book_list_unauthorized_view", args=(self.shelf_a.id,))
+        )
+        self.assertEqual(r.status_code, 401)
+
+    def test_unauthorized_user_failing(self):
+        del FilterByOrganization.permission_classes
+        del FilterByParent.permission_classes
```
nemesisdesign, line 192: I think I didn't explain myself properly. This is not required.
The test above is enough, we just have to ensure it fails without the changes to the code in the PR.

I tried this and it works, here's how to do it in this case:

- comment out `permission_classes` in the filter classes
- run tests
- expect the test above to fail with "AttributeError: 'AnonymousUser' object has no attribute 'organizations_dict'"

If it fails, it means the test is correct, if it doesn't it mean the test is missing something.

I confirm the test above is correct, this one below is not necessary and I'm removing it.

nemesisdesign (dismissed):
Thanks @Purhan, I'm pushing some improvements. See also my comments below.

PR #232, opened 2021-02-27
[chores] Improvement to admin test

Purhan:


Purhan:
Related: https://github.com/openwisp/openwisp-radius/pull/237

@nemesisdesign adding a new field to the users admin can change the order in which the permissions appear. Currently this will fail the integration tests for the PR mentioned above. I've switched it with a regex statement which will ensure it still passes. Please take a look.

openwisp_users/tests/test_admin.py
```diff
@@ -235,8 +236,8 @@ def test_admin_change_non_superuser_readonly_fields(self):
             reverse(f'admin:{self.app_label}_user_change', args=[operator.pk])
         )
         with self.subTest('User Permissions'):
-            html = f'<div class="readonly">{self.app_label}'
-            self.assertContains(response, html)
+            html = f'<div class="readonly">((?!</div>).)*({self.app_label})'
```
nemesisdesign, line 239: could you add a brief comment explaining how the regex works (what it matches)?

nemesisdesign (changes requested):
Thanks @Purhan see my small request below.

openwisp_users/tests/test_admin.py
Purhan: @nemesisdesign Currently the test checks in the response if `<div class="readonly">sample_users` exists.

After the change, the regex checks if `<div class="readonly">another_app | .... | sample_users` exists. This will also make sure that `sample_users` is inside the same html tag (that is, `<div class="readonly">` does not close with a `</div>` before it)

nemesisdesign (approved):
Thanks @Purhan

PR #199, opened 2020-11-15
[deps] Dropped Django 2.2 support

devkapilbansal:
Closes #180

PR #233, opened 2021-03-10
[fix] Fixed - Exception raised on opening non existing user ID #228

### LINKED ISSUE #228: [bug] Regression: opening non existing user ID raises an exception
How to replicate the issue:

**Try opening:**

- `http://localhost:8000/admin/openwisp_users/user/15f0345b-b9c1-44b2-b674-0deac5c7c816/change/`
- `http://localhost:8000/admin/openwisp_users/user/WRONG/change/`

**Expected outcome:** 404, no exception raised

**Actual outcome:** exception is raised.

**Cause:**

https://github.com/openwisp/openwisp-users/blob/8b4e84a665f0d53ddf3e96fb9a52d0641bd1e74b/openwisp_users/admin.py#L411

**Solution:**

- write a failing test for both cases
- handle both exceptions and raise Http404()

Saurav-Shrivastav:
- Added a test to check the response status on navigating to a non-existing user change page
- Handled exceptions for an invalid UUID and a non-existing user

Fixes #228

### DIFF
--- a/openwisp_users/admin.py
+++ b/openwisp_users/admin.py
@@ -409,8 +409,8 @@ class UserAdmin(MultitenantAdminMixin, BaseUserAdmin, BaseAdmin):
 
     def change_view(self, request, object_id, form_url='', extra_context=None):
         extra_context = extra_context or {}
-        obj = self.model.objects.get(pk=object_id)
-        if user_not_allowed_to_change_owner(request.user, obj):
+        obj = self.get_object(request, object_id)
+        if obj is not None and user_not_allowed_to_change_owner(request.user, obj):
             show_owner_warning = True
             extra_context.update({'show_owner_warning': show_owner_warning})
         return super().change_view(request, object_id, form_url, extra_context)
--- a/openwisp_users/tests/test_admin.py
+++ b/openwisp_users/tests/test_admin.py
@@ -2,6 +2,7 @@ import contextlib
 import os
 import re
 import smtplib
+import uuid
 from unittest.mock import patch
 
 from django.contrib import admin
@@ -168,6 +169,24 @@ class TestUsersAdmin(TestOrganizationMixin, TestUserAdditionalFieldsMixin, TestC
         self.assertEqual(queryset.count(), 0)
         self.assertEqual(len(mail.outbox), 0)
         self.assertContains(response, 'errors field-email')
+
+    def test_admin_change_user_page_get_invalid_UUID(self):
+        admin = self._create_admin()
+        self.client.force_login(admin)
+        with self.subTest('Test for wrong identifier'):
+            response = self.client.get(
+                reverse(f'admin:{self.app_label}_user_change', args=['WRONG']),
+                follow=True,
+            )
+            content = 'User with ID “WRONG” doesn’t exist. Perhaps it was deleted?'
+            self.assertContains(response, content, status_code=200)
+        with self.subTest('Test for non-existing user'):
+            id = uuid.uuid4()
+            response = self.client.get(
+                reverse(f'admin:{self.app_label}_user_change', args=[id],), follow=True
+            )
+            content = f'User with ID “{id}” doesn’t exist. Perhaps it was deleted?'
+            self.assertContains(response, content, status_code=200)
 
     def test_organization_view_on_site(self):
         admin = self._create_admin()


openwisp_users/tests/test_admin.py
```diff
@@ -169,6 +169,21 @@ def test_admin_change_user_email_empty(self):
...
+    def test_admin_change_user_page_get_invalid_UUID(self):
+        admin = self._create_admin()
+        self.client.force_login(admin)
+        response = self.client.get(
+            reverse(f'admin:{self.app_label}_user_change', args=['WRONG'])
+        )
+        response2 = self.client.get(
+            reverse(
+                f'admin:{self.app_label}_user_change',
+                args=['15f0345b-b9c1-44b2-b674-0deac5c7c816'],
```
pandafy, line 181:
```suggestion
                args=[uuid.uuid4()],
```
I believe instead of hard coding the identifier, using the `uuid` module to generate identifier will improve code readability.

openwisp_users/tests/test_admin.py
```diff
@@ -169,6 +169,21 @@ def test_admin_change_user_email_empty(self):
         self.assertEqual(len(mail.outbox), 0)
         self.assertContains(response, 'errors field-email')
 
+    def test_admin_change_user_page_get_invalid_UUID(self):
+        admin = self._create_admin()
+        self.client.force_login(admin)
+        response = self.client.get(
+            reverse(f'admin:{self.app_label}_user_change', args=['WRONG'])
+        )
```
pandafy, line 177: It is generally a good practice to keep testing code and assertions together. 

In this case, we can use [`subTest`](https://www.caktusgroup.com/blog/2017/05/29/subtests-are-best/) to further improve the readability. I would structure this test like below:

```python
with self.subTest('Test for wrong identifier'):
    response = self.client.get(
            reverse(f'admin:{self.app_label}_user_change', args=['WRONG'])
    )
    self.assertEquals(response.status_code, 404)
``` 

Similarly, you can add a subTest for the other case, `Test for non-existing user`.

pandafy (changes requested):
Good work @Saurav-Shrivastav ! :+1: 

I have requested some minor changes, see my comments below.

Saurav-Shrivastav:
Thanks for the suggestions @pandafy. I've implemented them.

openwisp_users/admin.py
```diff
@@ -409,7 +409,12 @@ def get_inline_instances(self, request, obj=None):
...
     def change_view(self, request, object_id, form_url='', extra_context=None):
         extra_context = extra_context or {}
-        obj = self.model.objects.get(pk=object_id)
+        try:
+            obj = self.model.objects.get(pk=object_id)
+        except self.model.DoesNotExist:
+            raise Http404("User Not Found.")
+        except ValidationError:
+            raise Http404("Invalid UUID")
         if user_not_allowed_to_change_owner(request.user, obj):
```
pandafy, line 418: To our ease, `django.contrib.admin.ModelAdmin` implements a `get_object` method that does error handling as well. Hence we can update the code as follows
```suggestion
        obj = self.get_object(request, object_id)
        if (obj is not None) and (user_not_allowed_to_change_owner(request.user, obj)):
```

openwisp_users/tests/test_admin.py
```diff
@@ -169,6 +170,20 @@ def test_admin_change_user_email_empty(self):
...
+            response = self.client.get(
+                reverse(f'admin:{self.app_label}_user_change', args=['WRONG'])
+            )
+            self.assertEquals(response.status_code, 404)
+        with self.subTest('Test for non-existing user'):
+            response2 = self.client.get(
+                reverse(f'admin:{self.app_label}_user_change', args=[uuid.uuid4()],)
+            )
+            self.assertEquals(response2.status_code, 404)
+
```
pandafy, line 186: These tests will be needed to be updated too. Since. the user will be redirected to the index admin page, you will need to pass `follow=True` in `self.client.get` method so it follows the redirect. 

The assertions will now check whether similar text exists in the response:
`User with ID “wrong” doesn’t exist. Perhaps it was deleted?`

pandafy (changes requested):
@Saurav-Shrivastav I manually tested the code today found a discrepancy. While you have achieved what was described in the issue description, that breaks the normal flow of Django Admin.

The `change_view` method of `django.contrib.admin.ModelAdmin` is implemented in a way that if an object does not exists or if validation fails, the user is redirected to the admin index page. 

See following screenshots which I took while testing similar behaviour with change_view of organization model.

image
image

With the current patch, the user will be greeted with a `404 Not Found` page if the corresponding object of `User` model does not exists or fails Validation.

I have shared suggestions to improve the current patch so it follows typical flow of Django Admin.

openwisp_users/admin.py
Saurav-Shrivastav: Made this change, works as expected now.

openwisp_users/tests/test_admin.py
Saurav-Shrivastav: There is an issue that I've been facing with the assertions in this case, it is because of the differences in the HTML response returned by Django 2.2 and Django 3.1. Django 2.2 returns the HTML in form of HTML entities, so a quote ` " ` is returned as `&quot;`, but it is in the form of a normal `"` in Django 3.1.

I tried to find a way around by decoding the HTML returned in Django 2.2 as:
```py
response.content = html.unescape(response.content.decode())
``` 
But it still doesn't seem to work.
Could you help me with it?

The subtest:
```py
response = self.client.get(
    reverse(f'admin:{self.app_label}_user_change', args=['WRONG']),
    follow=True,
)
content = 'User with ID'
response.content = html.unescape(response.content.decode())
self.assertIn(content, response.content.decode())
```

pandafy: @Shagunjain10 use [`assertContains`](https://docs.djangoproject.com/en/3.1/topics/testing/tools/#django.test.SimpleTestCase.assertContains) instead of `assertIn`. Read [Django testing documentation](https://docs.djangoproject.com/en/3.1/topics/testing/) to learn more writing tests.

Saurav-Shrivastav: You probably meant to tag me XD.
I've used `assertContains` too, but the actual problem is `“` and `"`. You see there is this difference between the quotes, that is why the assertion fails. So, it's majorly because, different Django versions return different HTML entities in the response. :/

Purhan: @Saurav-Shrivastav You can test for different django versions accordingly:

```python
if django.VERSION < (3, 1):
  # check for version < 3.1
else:
  # check for other versions
```

openwisp_users/admin.py
```diff
@@ -409,8 +409,8 @@ def get_inline_instances(self, request, obj=None):
 
     def change_view(self, request, object_id, form_url='', extra_context=None):
         extra_context = extra_context or {}
-        obj = self.model.objects.get(pk=object_id)
-        if user_not_allowed_to_change_owner(request.user, obj):
+        obj = self.get_object(request, object_id)
+        if (obj is not None) and (user_not_allowed_to_change_owner(request.user, obj)):
```
Purhan, line 413: You can probably drop some parentheses here

openwisp_users/tests/test_admin.py
Saurav-Shrivastav: I am thinking of using regex for the same, though I don't have much idea about it, will probably read about it. Do you think that it is a viable approach?

Saurav-Shrivastav: @Purhan , Won't that cause problems, when we shift to other versions of Django in future?

openwisp_users/admin.py
Saurav-Shrivastav: Yes, will do that.

openwisp_users/tests/test_admin.py
Purhan: > @Purhan , Won't that cause problems, when we shift to other versions of Django in future?

No, we can remove that when we drop support for some version, just make sure to leave a comment like this

```
#TODO: Remove this when we drop support for django<3.0
```

Saurav-Shrivastav: Thanks for the help @pandafy and @Purhan. The PR now looks complete. Please let me know, if I've missed out on anything.

openwisp_users/admin.py
```diff
@@ -409,8 +409,8 @@ def get_inline_instances(self, request, obj=None):
 
     def change_view(self, request, object_id, form_url='', extra_context=None):
         extra_context = extra_context or {}
-        obj = self.model.objects.get(pk=object_id)
-        if user_not_allowed_to_change_owner(request.user, obj):
+        obj = self.get_object(request, object_id)
+        if (obj is not None) and user_not_allowed_to_change_owner(request.user, obj):
```
nemesisdesign, line 413: the rounded parentheses are redundant, can you remove them please?

openwisp_users/tests/test_admin.py
```diff
@@ -169,6 +172,36 @@ def test_admin_change_user_email_empty(self):
...
+            id = uuid.uuid4()
+            response = self.client.get(
+                reverse(f'admin:{self.app_label}_user_change', args=[id],), follow=True
+            )
+            # TODO: Remove this when we drop support for django<3.0
+            if django.VERSION < (3, 1):
+                # check for version < 3.1
+                content = f'User with ID "{id}" doesn\'t exist. Perhaps it was deleted?'
+            else:
+                content = f'User with ID “{id}” doesn’t exist. Perhaps it was deleted?'
```
nemesisdesign, line 201: I would have avoided the version switch and checked only for part of the string, however, since it's already done and the check works, we can leave it, thanks.

openwisp_users/admin.py
nemesisdesign:
```suggestion
        if obj is not None and user_not_allowed_to_change_owner(request.user, obj):
```

nemesisdesign (changes requested):
Thanks to everyone involved! I spotted one aspect which can be improved.

openwisp_users/admin.py
Saurav-Shrivastav: Thanks for pointing it out, I've made the change. :+1:

openwisp_users/tests/test_admin.py
Saurav-Shrivastav: Okay!

Saurav-Shrivastav: Do you mean something like this? 
```py
content = '<li class="warning">User with ID '
```

openwisp_users/tests/test_admin.py
```diff
@@ -169,6 +172,36 @@ def test_admin_change_user_email_empty(self):
...
+            response.content = html.unescape(response.content.decode())
```
pandafy, line 202: You don't need to do this here. You can remove it.

pandafy (changes requested):

openwisp_users/tests/test_admin.py
Saurav-Shrivastav: Are you referring to line no 202?

pandafy: Yes!

Saurav-Shrivastav: It is required for Django 2.2, since it converts HTML entities like `&quot;` to `"`

nemesisdesign: Thanks @Saurav-Shrivastav, but support for django 2.2 was already planned to be dropped, the PR was ready but not merged yet, now is merged: https://github.com/openwisp/openwisp-users/pull/199.

nemesisdesign (changes requested):
Thanks @Saurav-Shrivastav, but support for django 2.2 was already planned to be dropped, the PR was ready but not merged yet, now is merged: https://github.com/openwisp/openwisp-users/pull/199.

Saurav-Shrivastav:
@nemesisdesign I have removed the portion of code that checked the django version, looks good now ^^

nemesisdesign (approved):

nemesisdesign:
Thanks @Saurav-Shrivastav :+1:

PR #234, opened 2021-03-24
[change] Make email lowercase when adding new user #227

### LINKED ISSUE #227: [change] Make email lowercase during registration
Let's avoid the possibility of having multiple users with the same email but with different cases for letters.

We should edit the register serializer to ensure the email is converted to lowercase (+ unit test).

Saurav-Shrivastav:
### Why:
[RFC 5321](https://tools.ietf.org/rfc/rfc5321.txt) states that the `user` in `user@hostname` of an email format is case-sensitive. `USER@example.com` and `user@example.com` are therefore different email addresses (the domain is case-insensitive).

### Changes:
- Overrode the save method of the `AbstractUser` model to make the email lowercase.
- Added a test to check that email is lowered.
- Added a test to ensure that an Integrity error is raised if a new user is created with an existing mail but different case.
- Modified `UserCreationForm`to check if the case-sensitive mail already exists.

Fixes #227

### DIFF
--- a/openwisp_users/base/models.py
+++ b/openwisp_users/base/models.py
@@ -3,6 +3,7 @@ import uuid
 
 import django
 from allauth.account.models import EmailAddress
+from django.contrib.auth import get_user_model
 from django.contrib.auth.models import AbstractUser as BaseUser
 from django.contrib.auth.models import UserManager as BaseUserManager
 from django.core.cache import cache
@@ -141,10 +142,21 @@ class AbstractUser(BaseUser):
         return self.__get_orgs('is_owner')
 
     def clean(self):
-        if self.email == '':
+        email = self.email
+        if email == '':
             self.email = None
         if self.phone_number == '':
             self.phone_number = None
+        if (
+            email
+            and get_user_model()
+            .objects.filter(email__iexact=email)
+            .exclude(pk=self.pk)
+            .exists()
+        ):
+            raise ValidationError(
+                {'email': _('User with this Email address already exists.')}
+            )
 
     @property
     def permissions(self):
--- a/openwisp_users/base/models.py
+++ b/openwisp_users/base/models.py
@@ -3,7 +3,6 @@ import uuid
 
 import django
 from allauth.account.models import EmailAddress
-from django.contrib.auth import get_user_model
 from django.contrib.auth.models import AbstractUser as BaseUser
 from django.contrib.auth.models import UserManager as BaseUserManager
 from django.core.cache import cache
@@ -142,15 +141,13 @@ class AbstractUser(BaseUser):
         return self.__get_orgs('is_owner')
 
     def clean(self):
-        email = self.email
-        if email == '':
+        if self.email == '':
             self.email = None
         if self.phone_number == '':
             self.phone_number = None
         if (
-            email
-            and get_user_model()
-            .objects.filter(email__iexact=email)
+            self.email
+            and self._meta.model.objects.filter(email__iexact=self.email)
             .exclude(pk=self.pk)
             .exists()
         ):
--- a/openwisp_users/tests/test_admin.py
+++ b/openwisp_users/tests/test_admin.py
@@ -460,6 +460,55 @@ class TestUsersAdmin(TestOrganizationMixin, TestUserAdditionalFieldsMixin, TestC
             res, '<li>User with this Email address already exists.</li>'
         )
 
+    def test_create_user_existing_mail_different_case(self):
+        self._create_user(username='user1', email='user@example.com')
+        admin = self._create_admin()
+        self.client.force_login(admin)
+        params = dict(
+            username='testadd',
+            email='USER@example.com',
+            password1='tester',
+            password2='tester',
+        )
+        params.update(self.add_user_inline_params)
+        res = self.client.post(reverse(f'admin:{self.app_label}_user_add'), params)
+        content = '<li>User with this Email address already exists.</li>'
+        self.assertContains(res, content, status_code=200)
+
+    def test_update_user_no_validation_error(self):
+        admin = self._create_admin()
+        self.client.force_login(admin)
+        user = self._create_user(email='user@example.com', username='user1')
+        params = user.__dict__
+        params['username'] = 'user2'
+        params.pop('last_login')
+        params.pop('phone_number')
+        params.pop('_password')
+        params = self._additional_params_pop(params)
+        params.update(self.add_user_inline_params)
+        params.update(
+            {
+                'emailaddress_set-TOTAL_FORMS': 1,
+                'emailaddress_set-INITIAL_FORMS': 1,
+                'emailaddress_set-0-verified': True,
+                'emailaddress_set-0-primary': True,
+                'emailaddress_set-0-id': user.emailaddress_set.first().id,
+                'emailaddress_set-0-user': user.id,
+            }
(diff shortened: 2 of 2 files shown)


openwisp_users/admin.py
```diff
@@ -185,6 +185,16 @@ class Meta(BaseUserCreationForm.Meta):
     class Media:
         js = ('admin/js/jquery.init.js', 'openwisp-users/js/addform.js')
 
+    def clean_email(self):
```
codesankalp, line 188: https://github.com/openwisp/openwisp-users/blob/114b16b51a415223cb2f34fd2a753bfe6a35d27f/openwisp_users/base/models.py#L50-L51

When the email is unique then it will automatically raise validation error @Saurav-Shrivastav.
I think it is not needed here.

Saurav-Shrivastav: Django doesn't check for the case of the email, that is why I've added the validation in the form. (unique=True, is case-insensitive). If a new user is created from the admin-panel - it'll lead to `Integrity Error`, in order to handle that gracefully, the `clean_email` function is required.

codesankalp: But in the `save` method you are making it lowercase. Then according to this, email with uppercase letters will always be converted into the lower case before saving, and when it will try to save pre-existing email it will raise validation error.
Correct me if I am wrong.

pandafy (commented):
I can't really tell if this is something that is required. An advantage of using a batteries included web framework like Django is that they already made some choices which we don't need to bothered about. 

I think before going ahead with this in OpenWISP, we should check why Django didn't implemented it as such in the first place. That'll require going through [Django's Trac](https://code.djangoproject.com/about) and checking relevant tickets. 

A quick search got me across the following similar tickets: 
- https://code.djangoproject.com/ticket/2273
- https://code.djangoproject.com/ticket/17561

I believe finding a reasoning for making such change will be a rabbit hole and will require a lot of time. 

On a personal note, to me this looks like more of a Django issue rather than an OpenWISP issue.

--------------------------------------

@nemesisdesign any quick thoughts on this?

openwisp_users/admin.py
Saurav-Shrivastav: When a request for adding a user is made, the validation takes place first and then the `save` method is reached at a later stage. So, the Validation at the form level will pass for the case insensitive email and it'll reach the `save` method, which will then throw an `IntegrityError: email already exists`.

openwisp_users/admin.py
```diff
@@ -185,6 +185,16 @@ class Meta(BaseUserCreationForm.Meta):
...
+        email = self.cleaned_data.get('email')
+
+        if (
+            email
+            and User.objects.filter(email__iexact=email).exclude(email=email).exists()
+        ):
+            raise forms.ValidationError('User with this Email address already exists.')
+        return email
```
nemesisdesign, line 196: if you follow my suggestion, this method won't be needed and you can remove it.

openwisp_users/tests/test_models.py
```diff
@@ -43,6 +44,19 @@ def test_create_user_without_email(self):
...
+        with self.subTest("Test that email is lowered"):
+            user = User.objects.get(username='user1')
+            self.assertEqual('user@example.com', user.email)
+        with self.subTest(
+            "Test creation of new user with the same mail but different case"
+        ):
+            u = User(username='user2', email='uSeR@example.com')
+            with self.assertRaises(Exception) as raised:
+                u.save()
+            self.assertEqual(IntegrityError, type(raised.exception))
```
nemesisdesign, line 58: we don't want to fail with IntegrityError but with ValidationError, which is catched by the django admin and shown to the user in a comprehensible way (IntegrityError would fail with a generic 500 internal server error).

nemesisdesign (changes requested):
The recommended fix is to use the clean() method. Since we may need to perform this cleanup before the unique validation (so django raises a ValidationError), overriding clean() may not be sufficient because as far as I remember it's executed after the unique fields validation. In these cases I override full_clean(), eg:

```
def full_clean(self, *args, **kwargs):
    # call custom method here, eg self._clean_email_case()
    return super().full_clean(*args, **kwargs)
```

nemesisdesign (changes requested):
**PS**: after reading through the material linked by @pandafy I changed a bit my perspective.

I think we can avoid the conversion of the email case to lowercase and we can simply add a custom check like:

```python
if User.objects.filter(email__iexact=email).exclude(pk=self.pk).exists():
    raise ValidationError({'email': _('Another user with this email address is already present.')})
```

Moreover, we can do this in `clean` rather than `full_clean` which will result in more linear code.

So the test can be changed to just check the validation error is raised correctly.

Saurav-Shrivastav:
I've made the suggested change @nemesisdesign @pandafy . :+1:

atb00ker (dismissed):

codesankalp (approved):

openwisp_users/admin.py
```diff
@@ -185,6 +185,15 @@ class Meta(BaseUserCreationForm.Meta):
     class Media:
         js = ('admin/js/jquery.init.js', 'openwisp-users/js/addform.js')
 
+    def clean(self):
```
nemesisdesign, line 188: I don't remember having asked that we do this in the admin form, why do this in the admin when we can do it in the main model? We always implement these kind of operations in the models, because this validation logic can be shared across different views (eg: admin, API views, custom views).

https://github.com/openwisp/openwisp-users/blob/master/openwisp_users/base/models.py#L143-L147

openwisp_users/admin.py
```diff
@@ -185,6 +185,15 @@ class Meta(BaseUserCreationForm.Meta):
...
+        data = super().clean()
+        email = data.get('email')
+        if email and User.objects.filter(email__iexact=email).exists():
```
nemesisdesign, line 191: you're forgetting to exclude the current user from the query. This means that validating the current user will fail. Please write a test for this case (ensuring that validating an existing user passes).

nemesisdesign (changes requested):
Not quite there yet. @atb00ker @codesankalp please take a look since you marked as approved.

openwisp_users/admin.py
Saurav-Shrivastav: I thought of implementing it "closer to the user". But you are right on this one, I will take care of this from next time. Will make the change soon.

openwisp_users/admin.py
Saurav-Shrivastav: Yes, since the code was implemented in the UserCreation Form, I didn't really think it was required, but I'll include it in the new patch.

nemesisdesign (commented):
Looks good, will test it asap.

openwisp_users/base/models.py
```diff
@@ -141,10 +142,21 @@ def organizations_owned(self):
         return self.__get_orgs('is_owner')
 
     def clean(self):
-        if self.email == '':
+        email = self.email
+        if email == '':
             self.email = None
```
devkapilbansal, line 147:
```suggestion
        if email == '':
            email = None
```

devkapilbansal (changes requested):
@Saurav-Shrivastav Looks good
Just one minor change here :point_down:

nemesisdesign (approved):
Thanks @Saurav-Shrivastav :+1:

PR #236, opened 2021-04-03
[fix] Fixed I/O on closed stream error

devkapilbansal:
Used redirect_any_output to remove noisy output instead of devnull.

### DIFF
--- a/openwisp_users/tests/test_admin.py
+++ b/openwisp_users/tests/test_admin.py
@@ -1,5 +1,4 @@
 import contextlib
-import os
 import re
 import smtplib
 import uuid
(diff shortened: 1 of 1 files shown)


openwisp_users/tests/test_admin.py
```diff
@@ -30,8 +32,6 @@
 User = get_user_model()
 Group = load_model('openwisp_users', 'Group')
 
-devnull = open(os.devnull, 'w')
-
```
devkapilbansal, line 34: Was opened outside of class but closed in `TearDown`. Therefore, it may cause issues if another tests are added that try to silence noise.

nemesisdesign (approved):
:+1:

PR #239, opened 2021-04-20
[fix] Restrict who can set objects to Organization=None #235

### LINKED ISSUE #235: [bug] DRF mixins which restrict who can set objects to organization=None (shared)
Make sure shared objects can be created and edited only by super administrators.
Non super administrators should only be able to view shared objects.

Otherwise operators with limited access to the system can create objects which can affect the entire system also for other organizations.

EG: the user of an org may create a required shared template which adds their SSH keys to the devices of all the organizations in the system, effectively gaining access to devices that they are not authorized to see.

I'm not sure how we could implement this, maybe we could implement this directly in `FilterSerializerByOrganization`?

What do you think @Purhan @ManishShah120?

BTW @Purhan, this is a question for you, are the children of `FilterSerializerByOrganization` supposed to be used always in serializers? Because if so, which I think it's the case, it would make the code more readable to move these classes in `serializers.py`.

ManishShah120:
Restrict Non-Super users from
setting objects to organization=None

Closes #235

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -655,7 +655,8 @@ without filtering by the organization the user has access to, which
 breaks multi-tenancy.
 
 The ``FilterSerializerByOrgMembership``, ``FilterSerializerByOrgManaged``
-and ``FilterSerializerByOrgOwned`` can be used to solve this issue.
+and ``FilterSerializerByOrgOwned`` can be used to solve this issue. These
+serializers obstructs non super-users from creating shared objects.
 
 Usage example:
 
--- a/README.rst
+++ b/README.rst
@@ -655,8 +655,9 @@ without filtering by the organization the user has access to, which
 breaks multi-tenancy.
 
 The ``FilterSerializerByOrgMembership``, ``FilterSerializerByOrgManaged``
-and ``FilterSerializerByOrgOwned`` can be used to solve this issue. These
-serializers obstructs non super-users from creating shared objects.
+and ``FilterSerializerByOrgOwned`` can be used to solve this issue.
+
+These serializers do not allow non-superusers to create shared objects.
 
 Usage example:
 
--- a/tests/testapp/tests/test_filter_classes.py
+++ b/tests/testapp/tests/test_filter_classes.py
@@ -185,3 +185,14 @@ class TestFilterClasses(TestMultitenancyMixin, TestCase):
             reverse("test_book_list_unauthorized_view", args=(self.shelf_a.id,))
         )
         self.assertEqual(r.status_code, 401)
+
+    def test_presence_of_null_org_field(self):
+        operator = self._get_operator()
+        self._create_org_user(
+            user=operator, is_admin=True, organization=self._get_org('org_a')
+        )
+        token = self._obtain_auth_token(operator)
+        auth = dict(HTTP_AUTHORIZATION=f'Bearer {token}')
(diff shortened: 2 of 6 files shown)


Purhan (commented):
LGTM :+1: but would be good to add a test as well.

nemesisdesign (changes requested):
You see? What we needed could be implemented in 1 line.
This is what we have to strive for, achieve what we need with simplicity.

@Purhan is right, but I also think there are a few other things missing, so please add:

- one test, you could send a request to an API endpoint which has this serializer, using content-type text/HTML, which should give you a response from the browsable API, and there you can test for the presence of the null option, please make sure the test fails without your change and passes with it
- but also a line of doc in https://github.com/openwisp/openwisp-users#multi-tenant-serializers-for-the-browsable-web-ui which explains that the serializers are designed to not allow non-superadmin users to create shared objects
- in case of superusers, it would be great to show the same label for the null option as we do in the admin, if you can do that easily please try

ManishShah120:
Hi, @nemesisdesign , @atb00ker , @Purhan 
- All the present API endpoints present in the `testapp` didn't allowed any shared objects creation for any model, so could not write tests with the present endpoints. To test this I had to introduce one extra endpoint for the `Template` Model, which allowed shared object creation.
- Have updated the docs.

tests/testapp/tests/test_filter_classes.py
```diff
@@ -185,3 +185,14 @@ def test_unauthorized_user(self):
...
+
+    def test_presence_of_null_org_field(self):
+        operator = self._get_operator()
+        self._create_org_user(
+            user=operator, is_admin=True, organization=self._get_org('org_a')
+        )
+        token = self._obtain_auth_token(operator)
+        auth = dict(HTTP_AUTHORIZATION=f'Bearer {token}')
+        path = reverse('test_template_list') + '?format=api'
+        response = self.client.get(path, **auth)
```
nemesisdesign, line 197: this is a lot cleaner:
```python
response = self.client.get(url, {'format': 'api'}, HTTP_AUTHORIZATION=f'Bearer {token}')
```

I noticed only now the same pattern in the rest of the file, I will open an issue shortly to clean that up later for now just clean this one please.

README.rst
```diff
@@ -655,7 +655,8 @@ without filtering by the organization the user has access to, which
 breaks multi-tenancy.
 
 The ``FilterSerializerByOrgMembership``, ``FilterSerializerByOrgManaged``
-and ``FilterSerializerByOrgOwned`` can be used to solve this issue.
+and ``FilterSerializerByOrgOwned`` can be used to solve this issue. These
+serializers obstructs non super-users from creating shared objects.
```
nemesisdesign, line 659: Please move the new sentence on its own line and let's change it to:

"These serializers do not allow non-superusers to create shared objects".

nemesisdesign (changes requested):
Great work @ManishShah120, see my comments below for minor cleanups and we'll be ready to merge.

tests/testapp/tests/test_filter_classes.py
nemesisdesign: Created issue: https://github.com/openwisp/openwisp-users/issues/242

nemesisdesign (approved):
Thanks!

PR #248, opened 2021-05-06
[fix] Include shared objects while filtering serilaizer fields #247

### LINKED ISSUE #247: [Bug] Include the objects with org=None, while filtering serilaizer fileds based on the organization of an user
https://github.com/openwisp/openwisp-users/blob/747ac34ac59c7fee843f8ad250cc5d12292c722b/openwisp_users/api/mixins.py#L137-L140

@nemesisdesign , @atb00ker with this line of code we are trying to filter out other `ForeignKey` related field other than `organization` field, based on the organization of a user, But mostly all of such fields has an option to create objects as shared org(organization=None). eg.:- templates, CA, etc.. But this LOC only returns the object belonging to an organization, excluding the objects whose organization is set to None.

Will it not be better, if we include the shared objects along with these.

CC: @Purhan

ManishShah120:
Closes #247

### DIFF
--- a/openwisp_users/api/mixins.py
+++ b/openwisp_users/api/mixins.py
@@ -118,7 +118,7 @@ class FilterSerializerByOrganization:
     """
 
     organization_lookup = 'organization__in'
-    filter_shared = True
+    filter_shared = False
 
     @property
     def _user_attr(self):
--- a/openwisp_users/api/mixins.py
+++ b/openwisp_users/api/mixins.py
@@ -1,5 +1,6 @@
 import swapper
 from django.core.exceptions import ValidationError
+from django.db.models import Q
 from rest_framework.exceptions import NotFound
 from rest_framework.permissions import IsAuthenticated
 
@@ -117,6 +118,7 @@ class FilterSerializerByOrganization:
     """
 
     organization_lookup = 'organization__in'
+    filter_shared = True
 
     @property
     def _user_attr(self):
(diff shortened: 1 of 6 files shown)


ManishShah120:
@nemesisdesign , @atb00ker In the testapp no instance with relating was allowed to be created with `org=None`, so for writting tests for these changes I had to allow `Shelf` instances to be created with `org=None`, so therefore, I edited the existing migration file in the testapp.

openwisp_users/api/mixins.py
```diff
@@ -117,6 +118,7 @@ class FilterSerializerByOrganization:
     """
 
     organization_lookup = 'organization__in'
+    filter_shared = True
```
nemesisdesign, line 121: let's set it to `False` by default please, so we have to enable it explicitly on models which inherit `ShareableOrgMixin`

nemesisdesign (changes requested):
Great progress Manish but seems to me you forgot to update the docs.

Please always make a habit to update documentation when already documented features are changed.

The docs for these serializers are here: https://github.com/openwisp/openwisp-users#multi-tenant-serializers-for-the-browsable-web-ui

Just mention this new attribute and provide an example.

See also my comment below.

openwisp_users/api/mixins.py
```diff
@@ -135,8 +137,11 @@ def filter_fields(self):
                 )
                 continue
             try:
+                conditions = Q(**{self.organization_lookup: organization_filter})
```
atb00ker, line 140: `conditions`, can we change the varible name to something like `org_filter` or `org_filter_condiitions` or something like that, `conditions` alone sounds out of place!

atb00ker (commented):
Looks good!

openwisp_users/api/mixins.py
nemesisdesign: `conditions` is fine, those are the conditions of the filter so it's good enough for me

nemesisdesign (approved):

PR #251, opened 2021-05-22
[feature] Implemented view permissions by extending `DjangoModelPermissions` class #249

### LINKED ISSUE #249: [feature] Add class to implement view permissions in DRF
DRF DjangoModelPermission does not takes into account the view permissions, and hence any user is allowed to view the objects even though they may not have view permission or change permission on that specific object.

We recently dealt with this in openwisp-network-topology, we could use a similar solution, adapted to be more generic:
https://github.com/openwisp/openwisp-network-topology/blob/3c3404748a9ecab65f3f4a389f94d2c75fbde732/openwisp_network_topology/api/views.py#L31-L51.

I think we could extend `DjangoModelPermissions` and add our logic so that:

- if user is superuser, any request is allowed
- if user is not superuser and has view or change permission, GET requests are allowed
- if user is not superuser and does not have view or change permission, GET requests are not allowed 

Here's a few possible tests explained below.

**Preconditions:**

- given the operator group does not have view permission to a specific model
- given the administrator group has change permission to the same model (view permission is omitted, ensure is missing)
- given two API endpoints for this model: list and detail

**Tests**:

- ensure an user of the operator group (not flagged as supersuser) **cannot**  view both endpoints 
- ensure an user of the administrator group (not flagged as supersuser) **can** view both endpoints 
- add the view permission to the operator group, ensure the user now can view both endpoints but cannot change anything (put or delete)

ManishShah120:
Closes #249

### DIFF
--- a/openwisp_users/api/permissions.py
+++ b/openwisp_users/api/permissions.py
@@ -1,5 +1,8 @@
 from django.utils.translation import gettext_lazy as _
-from rest_framework.permissions import BasePermission, DjangoModelPermissions
+from rest_framework.permissions import BasePermission
+from rest_framework.permissions import (
+    DjangoModelPermissions as BaseDjangoModelPermissions,
+)
 from swapper import load_model
 
 Organization = load_model('openwisp_users', 'Organization')
@@ -68,7 +71,7 @@ class IsOrganizationOwner(BaseOrganizationPermission):
         return org and (user.is_superuser or user.is_owner(org))
 
 
-class ViewDjangoModelPermissions(DjangoModelPermissions):
+class DjangoModelPermissions(BaseDjangoModelPermissions):
     perms_map = {
         'GET': ['%(app_label)s.view_%(model_name)s'],
         'OPTIONS': [],
@@ -94,9 +97,6 @@ class ViewDjangoModelPermissions(DjangoModelPermissions):
         change_perm = self.get_required_permissions('PUT', queryset.model)
 
         if request.method == 'GET':
-            if user.has_perms(perms) or user.has_perms(change_perm):
-                return True
-            else:
-                return False
+            return user.has_perms(perms) or user.has_perms(change_perm)
 
         return user.has_perms(perms)
--- a/openwisp_users/api/permissions.py
+++ b/openwisp_users/api/permissions.py
@@ -70,7 +70,7 @@ class IsOrganizationOwner(BaseOrganizationPermission):
         return org and (user.is_superuser or user.is_owner(org))
 
 
-class CustomDjangoModelPermissions(DjangoModelPermissions):
+class ViewDjangoModelPermissions(DjangoModelPermissions):
     def __init__(self):
         self.perms_map = copy.deepcopy(self.perms_map)
         self.perms_map['GET'] = ['%(app_label)s.view_%(model_name)s']
@@ -81,9 +81,8 @@ class CustomDjangoModelPermissions(DjangoModelPermissions):
         if getattr(view, '_ignore_model_permissions', False):
             return True
 
-        if not request.user or (
-            not request.user.is_authenticated and self.authenticated_users_only
-        ):
+        user = request.user
+        if not user or (not user.is_authenticated and self.authenticated_users_only):
             return False
 
         queryset = self._queryset(view)
@@ -91,9 +90,9 @@ class CustomDjangoModelPermissions(DjangoModelPermissions):
         change_perm = self.get_required_permissions('PUT', queryset.model)
 
         if request.method == 'GET':
-            if request.user.has_perms(perms) or request.user.has_perms(change_perm):
+            if user.has_perms(perms) or user.has_perms(change_perm):
                 return True
             else:
                 return False
 
-        return request.user.has_perms(perms)
+        return user.has_perms(perms)
--- a/openwisp_users/api/permissions.py
+++ b/openwisp_users/api/permissions.py
@@ -1,5 +1,7 @@
+import copy
+
 from django.utils.translation import gettext_lazy as _
-from rest_framework.permissions import BasePermission
+from rest_framework.permissions import BasePermission, DjangoModelPermissions
 from swapper import load_model
 
 Organization = load_model('openwisp_users', 'Organization')
@@ -66,3 +68,32 @@ class IsOrganizationOwner(BaseOrganizationPermission):
 
     def validate_membership(self, user, org):
         return org and (user.is_superuser or user.is_owner(org))
+
+
+class CustomDjangoModelPermissions(DjangoModelPermissions):
+    def __init__(self):
+        self.perms_map = copy.deepcopy(self.perms_map)
+        self.perms_map['GET'] = ['%(app_label)s.view_%(model_name)s']
+
+    def has_permission(self, request, view):
+        # Workaround to ensure DjangoModelPermissions are not applied
+        # to the root view when using DefaultRouter.
+        if getattr(view, '_ignore_model_permissions', False):
+            return True
+
+        if not request.user or (
+            not request.user.is_authenticated and self.authenticated_users_only
+        ):
+            return False
+
+        queryset = self._queryset(view)
+        perms = self.get_required_permissions(request.method, queryset.model)
+        change_perm = self.get_required_permissions('PUT', queryset.model)
+
+        if request.method == 'GET':
+            if request.user.has_perms(perms) or request.user.has_perms(change_perm):
+                return True
+            else:
+                return False
+
+        return request.user.has_perms(perms)
--- a/openwisp_users/api/permissions.py
+++ b/openwisp_users/api/permissions.py
@@ -1,5 +1,3 @@
-import copy
-
 from django.utils.translation import gettext_lazy as _
 from rest_framework.permissions import BasePermission, DjangoModelPermissions
 from swapper import load_model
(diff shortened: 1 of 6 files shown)


openwisp_users/api/permissions.py
```diff
@@ -66,3 +68,32 @@ class IsOrganizationOwner(BaseOrganizationPermission):
...
+
+        queryset = self._queryset(view)
+        perms = self.get_required_permissions(request.method, queryset.model)
+        change_perm = self.get_required_permissions('PUT', queryset.model)
+
+        if request.method == 'GET':
+            if request.user.has_perms(perms) or request.user.has_perms(change_perm):
+                return True
+            else:
+                return False
```
ManishShah120, line 97: Did not handled the case of super user becasue superuser by default has all the permissions.

tests/testapp/tests/test_permission_classes.py
```diff
@@ -117,3 +121,29 @@ def test_organization_field_with_errored_parent(self):
         with self.assertRaises(AttributeError) as error:
             self.client.get(reverse('test_error_field_view'), **auth)
         self.assertIn('Organization not found', str(error.exception))
+
+    def test_custom_django_model_permission_with_view_permission(self):
```
atb00ker, line 125: Can we have a test for negative case when user does not have the permission; they are given 403.

openwisp_users/api/permissions.py
```diff
@@ -66,3 +68,32 @@ class IsOrganizationOwner(BaseOrganizationPermission):
 
     def validate_membership(self, user, org):
         return org and (user.is_superuser or user.is_owner(org))
+
+
+class CustomDjangoModelPermissions(DjangoModelPermissions):
```
atb00ker, line 73: Just wondering, can we name it something like `ViewDjangoModelPermissions` so that the name makes more sense in the other modules about that's "Custom" about this class!

openwisp_users/api/permissions.py
atb00ker: I think you can add a check at the top stating:
```
if user.is_superuser:
    return True
```

They are allowed to do anything regardless of any permission!

atb00ker (commented):
This looks good to me, small comments:

openwisp_users/api/permissions.py
ManishShah120: @atb00ker I am trying to implement this in more generic way, cause am thinking of opening a similar PR in DRF repo once this gets merged here. 

By default `superuser` has all permissions i.e., it can do anything it wants. suppose manually someone removes all the permissions for any specific model for `superuser`. Then in that case this would not work, since we will return True if it;s a super user, so that case will fail.

What do you think?

openwisp_users/api/permissions.py
ManishShah120: Done @atb00ker :+1:

tests/testapp/tests/test_permission_classes.py
ManishShah120: I have updated the tests. :+1:

ManishShah120:
Quoting from the issue #249

> **Tests:**

> - ensure an user of the operator group (not flagged as supersuser) cannot view both endpoints
  - Added
> - ensure an user of the administrator group (not flagged as supersuser) can view both endpoints
  - Added
> - add the view permission to the operator group, ensure the user now can view both endpoints but cannot change anything (put or delete)
  - Added

openwisp_users/api/permissions.py
atb00ker: To the best of my knowledge, removing permissions from `superuser` is an anti-pattern, but I understand your point.
1. We should look at existing DRF functions and implement it in a similar way.
2. I'll also wait for @nemesisdesign check this! :smile:

openwisp_users/api/permissions.py
```diff
@@ -66,3 +68,31 @@ class IsOrganizationOwner(BaseOrganizationPermission):
 
     def validate_membership(self, user, org):
         return org and (user.is_superuser or user.is_owner(org))
+
+
+class ViewDjangoModelPermissions(DjangoModelPermissions):
+    def __init__(self):
+        self.perms_map = copy.deepcopy(self.perms_map)
+        self.perms_map['GET'] = ['%(app_label)s.view_%(model_name)s']
```
nemesisdesign, line 76: can we do this at class level so it's done only once at startup?

nemesisdesign (commented):
Looks on the right track! I'll need to test it.

But definitely please go ahead with the PR to Django REST Framework, hopefully we can get something similar or better merged there.

okraits (commented):
Looks fine. No objections from me.

openwisp_users/api/permissions.py
ManishShah120: Done

okraits (commented):
Looks fine.

tests/testapp/tests/test_permission_classes.py
```diff
@@ -117,3 +126,128 @@ def test_organization_field_with_errored_parent(self):
         with self.assertRaises(AttributeError) as error:
             self.client.get(reverse('test_error_field_view'), **auth)
         self.assertIn('Organization not found', str(error.exception))
+
+    def test_view_permission_with_operator(self):
+        user = User.objects.create_user(
+            username='operator', password='tester', email='operator@test.com'
+        )
+        operator_group = Group.objects.filter(name='Operator')
+        user.groups.set(operator_group)
```
atb00ker, line 135: Just FYI, I think there are functions for common actions. Like [_create_operator](https://github.com/openwisp/openwisp-users/blob/master/openwisp_users/tests/utils.py#L53).
These should be inherited and used! :smile:

tests/testapp/tests/test_permission_classes.py
```diff
@@ -117,3 +126,128 @@ def test_organization_field_with_errored_parent(self):
...
+        org1 = self._get_org()
+        OrganizationUser.objects.create(user=user, organization=org1, is_admin=True)
```
atb00ker, line 137: Same here. Find function here: https://github.com/openwisp/openwisp-users/blob/master/openwisp_users/tests/utils.py#L128

atb00ker (commented):
I think this is looking good, small improvement in the tests.

atb00ker (dismissed):
Functional Test:
- [x] superuser has access
- [x] non-su without view permission & org manager doesn't have access
- [x] non-su with view permission & not org manager has access but sees nothing
- [x] non-su with view permission & org manager can see temp for their obj. (not shared obj)

tests/testapp/tests/test_permission_classes.py
ManishShah120: Hey, @atb00ker Yes, I tried to use it but with this function, the user gets all the permissions without being in the `operator` group, and here we only needed `view.testapp_template` permission.

openwisp_users/api/permissions.py
```diff
@@ -66,3 +66,37 @@ class IsOrganizationOwner(BaseOrganizationPermission):
...
+
+        queryset = self._queryset(view)
+        perms = self.get_required_permissions(request.method, queryset.model)
+        change_perm = self.get_required_permissions('PUT', queryset.model)
+
+        if request.method == 'GET':
+            if user.has_perms(perms) or user.has_perms(change_perm):
+                return True
+            else:
+                return False
```
nemesisdesign, line 100: what about:
```python
# user must have either view permission or change permission
return user.has_perms(perms) or user.has_perms(change_perm)
```

tests/testapp/tests/test_permission_classes.py
```diff
@@ -117,3 +126,128 @@ def test_organization_field_with_errored_parent(self):
...
+        )
+        user_permissions = Permission.objects.filter(codename='change_template')
+        user.user_permissions.add(*user_permissions)
+        user.organizations_dict  # force caching
+        org1 = self._get_org()
+        OrganizationUser.objects.create(user=user, organization=org1, is_admin=True)
+        self.client.force_login(user)
+        token = self._obtain_auth_token()
+        auth = dict(HTTP_AUTHORIZATION=f'Bearer {token}')
+        t1 = self._create_template(organization=org1)
```
nemesisdesign, line 245: the first lines of each tests look really similar, please try to make a private helper method which prepares the data needed for the test, so we can make the actual test code shorter and easier to read.

openwisp_users/api/permissions.py
```diff
@@ -66,3 +66,37 @@ class IsOrganizationOwner(BaseOrganizationPermission):
 
     def validate_membership(self, user, org):
         return org and (user.is_superuser or user.is_owner(org))
+
+
+class ViewDjangoModelPermissions(DjangoModelPermissions):
```
nemesisdesign, line 71: we could call this simply `DjangoModelPermissions`

openwisp_users/api/permissions.py
```diff
@@ -1,5 +1,5 @@
 from django.utils.translation import gettext_lazy as _
-from rest_framework.permissions import BasePermission
+from rest_framework.permissions import BasePermission, DjangoModelPermissions
```
nemesisdesign, line 2: here you'd import it as `BaseDjangoModelPermissions`

nemesisdesign (changes requested):
Great work @ManishShah120 @atb00ker! :+1: 

Missing:

- [ ] Documentation (please add it near to the other sections related to DRF)
- [x] Code simplification (see below)

tests/testapp/tests/test_permission_classes.py
ManishShah120: Done, Please have a look. :+1:

tests/testapp/tests/test_permission_classes.py
```diff
@@ -117,3 +126,106 @@ def test_organization_field_with_errored_parent(self):
...
+        self.client.force_login(user)
+        token = self._obtain_auth_token(user)
+        auth = dict(HTTP_AUTHORIZATION=f'Bearer {token}')
+        t1 = self._create_template(organization=org1)
+        return (auth, t1)
+
+    def test_view_permission_with_operator(self):
+        user = self._get_user()
+        operator_group = Group.objects.filter(name='Operator')
+        user.groups.set(operator_group)
```
atb00ker, line 141: Doesn't `_get_operator()` help here?

atb00ker (approved):
LGTM! :+1:

tests/testapp/tests/test_permission_classes.py
ManishShah120: No @atb00ker.

nemesisdesign (approved):
:+1:

PR #258, opened 2021-06-11
Improve api mixins for organization field lookup in parent model #255

### LINKED ISSUE #255: [bug] API mixins don't take into account organization_lookup
Some classes in openwisp-users were not taking into account the organization_lookup field, hence queries were failing to find the right organization field, I solved this in https://github.com/openwisp/openwisp-users/tree/improve-api-mixins, but tests are missing, we need at least one test in openwisp-users.

I discovered this while working on https://github.com/openwisp/openwisp-firmware-upgrader/issues/142.

ManishShah120:
Closes #255


ManishShah120:
Not sure if any documentation update is required here :thinking:

tests/testapp/models.py
```diff
@@ -51,3 +51,12 @@ def __str__(self):
 
     class Meta:
         abstract = False
+
+
+class Library(models.Model):
```
atb00ker, line 56: Hi,

It's not clear to me, why did we need a new model to test this, was there a challenge with updating the old models?

atb00ker:
> Not sure if any documentation update is required here thinking

To be sure of this, think about:

1. User docs: Can this be useful for a user for setup / customization?
2. Developer docs: Can another module inherit the class that you created / modified? (like can this be used in openwisp-controller, if yes, we need to document how)
3. Is it hard to understand without documentation? (eg. repository setup)

If any of them are true, I think we should include information for it in docs! Does this help? 😄

tests/testapp/models.py
ManishShah120: @atb00ker we needed a model which does not contain the organization field, but it's parent model contains it. I thought it will be better to introduce new model, as all the present models contained organization field, and editing the already present model would fail some of the test cases.

nemesisdesign (commented):
We could mention the `organization_field` in every class that it's used and link to https://github.com/openwisp/openwisp-users#organization_field.

README.rst
```diff
@@ -748,6 +748,12 @@ Usage example:
             model = Book
             fields = '__all__'
 
+To filter items based on the ``organization`` of their parent object,
+``organization_field`` attribute can be defined in the view function
+which is inheriting any of the mixin classes.
+
+Usage example: `organization_fieled <https://github.com/openwisp/openwisp-users#organization_field>`_.
```
okraits, line 755: A little typo here :grinning:

okraits (commented):
Other than that - good so far.

atb00ker (approved):
LGTM! :smile:

nemesisdesign (approved):
:+1:

PR #260, opened 2021-06-12
[fix] Fixed `KeyError` in `FilterSerializerByOrgManaged` #259

### LINKED ISSUE #259: [mixin] `FilterSerializerByOrgManaged` raises `KeyError` when implemented to nested serializer classes
The `FilterSerializerByOrgManaged` class raises `KeyError` when we run the server after implementing in some serializer classes which contains nested serializer class.
I found this issue while working on :- https://github.com/openwisp/openwisp-firmware-upgrader/issues/142

ManishShah120:
Closes #259

### DIFF
--- a/openwisp_users/api/mixins.py
+++ b/openwisp_users/api/mixins.py
@@ -146,10 +146,10 @@ class FilterSerializerByOrganization(OrgLookup):
                     pk__in=organization_filter
                 )
                 continue
+            conditions = Q(**{self.organization_lookup: organization_filter})
+            if self.include_shared:
+                conditions |= Q(organization__isnull=True)
             try:
-                conditions = Q(**{self.organization_lookup: organization_filter})
-                if self.include_shared:
-                    conditions |= Q(organization__isnull=True)
                 self.fields[field].queryset = self.fields[field].queryset.filter(
                     conditions
                 )
--- a/openwisp_users/api/mixins.py
+++ b/openwisp_users/api/mixins.py
@@ -131,26 +131,29 @@ class FilterSerializerByOrganization(OrgLookup):
         raise NotImplementedError()
 
     def filter_fields(self):
-        user = self.context['request'].user
-        if user.is_superuser:
-            return
-        organization_filter = getattr(user, self._user_attr)
-        for field in self.fields:
-            if field == 'organization':
-                self.fields[field].allow_null = False
-                self.fields[field].queryset = self.fields[field].queryset.filter(
-                    pk__in=organization_filter
-                )
-                continue
-            try:
-                conditions = Q(**{self.organization_lookup: organization_filter})
-                if self.include_shared:
-                    conditions |= Q(organization__isnull=True)
-                self.fields[field].queryset = self.fields[field].queryset.filter(
-                    conditions
-                )
-            except AttributeError:
-                pass
+        try:
+            user = self.context['request'].user
+            if user.is_superuser:
+                return
+            organization_filter = getattr(user, self._user_attr)
+            for field in self.fields:
+                if field == 'organization':
+                    self.fields[field].allow_null = False
+                    self.fields[field].queryset = self.fields[field].queryset.filter(
+                        pk__in=organization_filter
+                    )
+                    continue
+                try:
+                    conditions = Q(**{self.organization_lookup: organization_filter})
+                    if self.include_shared:
+                        conditions |= Q(organization__isnull=True)
+                    self.fields[field].queryset = self.fields[field].queryset.filter(
+                        conditions
+                    )
+                except AttributeError:
+                    pass
+        except KeyError:
+            pass
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
--- a/openwisp_users/api/mixins.py
+++ b/openwisp_users/api/mixins.py
@@ -131,13 +131,11 @@ class FilterSerializerByOrganization(OrgLookup):
         raise NotImplementedError()
 
     def filter_fields(self):
-        if self.context:
-            user = self.context['request'].user
-        else:
-            return
-
+        user = self.context['request'].user
+        # superuser can see everything
         if user.is_superuser:
             return
+        # non superusers can see only items of organizations they're related to
         organization_filter = getattr(user, self._user_attr)
         for field in self.fields:
             if field == 'organization':
@@ -158,7 +156,10 @@ class FilterSerializerByOrganization(OrgLookup):
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
-        self.filter_fields()
+        # only filter related fields if the serializer
+        # is being initiated during an HTTP request
+        if 'request' in self.context:
+            self.filter_fields()
 
 
 class FilterSerializerByOrgMembership(FilterSerializerByOrganization):
--- a/openwisp_users/api/mixins.py
+++ b/openwisp_users/api/mixins.py
@@ -131,9 +131,9 @@ class FilterSerializerByOrganization(OrgLookup):
         raise NotImplementedError()
 
     def filter_fields(self):
-        try:
+        if self.context:
             user = self.context['request'].user
-        except KeyError:
+        else:
             return
 
         if user.is_superuser:
--- a/openwisp_users/api/mixins.py
+++ b/openwisp_users/api/mixins.py
@@ -133,27 +133,28 @@ class FilterSerializerByOrganization(OrgLookup):
     def filter_fields(self):
         try:
             user = self.context['request'].user
-            if user.is_superuser:
-                return
-            organization_filter = getattr(user, self._user_attr)
-            for field in self.fields:
-                if field == 'organization':
-                    self.fields[field].allow_null = False
-                    self.fields[field].queryset = self.fields[field].queryset.filter(
-                        pk__in=organization_filter
-                    )
-                    continue
-                try:
-                    conditions = Q(**{self.organization_lookup: organization_filter})
-                    if self.include_shared:
-                        conditions |= Q(organization__isnull=True)
-                    self.fields[field].queryset = self.fields[field].queryset.filter(
-                        conditions
-                    )
-                except AttributeError:
-                    pass
         except KeyError:
-            pass
+            return
+
+        if user.is_superuser:
+            return
+        organization_filter = getattr(user, self._user_attr)
+        for field in self.fields:
+            if field == 'organization':
+                self.fields[field].allow_null = False
+                self.fields[field].queryset = self.fields[field].queryset.filter(
+                    pk__in=organization_filter
+                )
+                continue
+            try:
+                conditions = Q(**{self.organization_lookup: organization_filter})
+                if self.include_shared:
+                    conditions |= Q(organization__isnull=True)
+                self.fields[field].queryset = self.fields[field].queryset.filter(
+                    conditions
+                )
+            except AttributeError:
+                pass
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
--- a/tests/testapp/serializers.py
+++ b/tests/testapp/serializers.py
@@ -61,15 +61,25 @@ class LibrarySerializer(FilterSerializerByOrgManaged, ValidatedModelSerializer):
 class ShelfSerializerForBook(FilterSerializerByOrgManaged, ValidatedModelSerializer):
     class Meta:
         model = Shelf
-        fields = ['name']
+        fields = '__all__'
+        read_only_fields = ('created', 'modified')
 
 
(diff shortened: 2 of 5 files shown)


openwisp_users/api/mixins.py
```diff
@@ -125,26 +125,29 @@ def _user_attr(self):
...
-            try:
-                conditions = Q(**{self.organization_lookup: organization_filter})
-                if self.include_shared:
-                    conditions |= Q(organization__isnull=True)
-                self.fields[field].queryset = self.fields[field].queryset.filter(
-                    conditions
-                )
-            except AttributeError:
-                pass
+        try:
```
nemesisdesign, line 128: never wrap an entire block in a try statement, only wrap the line which you expect can fail, otherwise we risk catching other errors as well but we won't understand what's going on and will be hard to debug.

nemesisdesign (changes requested):

openwisp_users/api/mixins.py
ManishShah120: @nemesisdesign the error is raised due to this line:- 
`user = self.context['request'].user`
I tried wrapping only this line in `try` statement but then other variables like `organization_filter` depends on it with which entire logic is dependent, it raises variable used before assignment error. 

I tried assigning None to it, but it didn't work.

atb00ker: Indeed, too many folks add all the code in the try block, we should avoid this.

> tried wrapping only this line in try statement but then other variables like organization_filter depends on it with which entire logic is dependent

Why not `return` in the catch block when getting user is failed?
Wouldn't that solve the problem?

ManishShah120: Done @atb00ker

openwisp_users/api/mixins.py
```diff
@@ -127,27 +127,28 @@ def _user_attr(self):
...
+                    pk__in=organization_filter
+                )
+                continue
+            try:
+                conditions = Q(**{self.organization_lookup: organization_filter})
+                if self.include_shared:
+                    conditions |= Q(organization__isnull=True)
+                self.fields[field].queryset = self.fields[field].queryset.filter(
+                    conditions
+                )
```
atb00ker: There is a `try` block here too, do all the lines need to be inside this block?

ManishShah120: @atb00ker the error is raised in by line 4
``` python
             conditions = Q(**{self.organization_lookup: organization_filter})
             if self.include_shared:                    
             conditions |= Q(organization__isnull=True)                
             self.fields[field].queryset = self.fields[field].queryset.filter(conditions)
```
Should we only that line in the try statement?

atb00ker: @ManishShah120 The question here https://github.com/openwisp/openwisp-users/pull/260#discussion_r650861223
Is repeated to ask the same thing answered here: https://github.com/openwisp/openwisp-users/pull/260#discussion_r650404300

So could you please elaborate the source of the confusion so that I don't give the same answer again! :smile:

openwisp_users/api/mixins.py
nemesisdesign: @ManishShah120 why would the request be missing from the serializer context?
Is there a legitimate reason for this to happen?
I think we should fail the request if `KeyError` is raised, because it means the application is misconfigured.
The serializer shall always have the request passed to the serializer context during initialization or otherwise it won't work.

Don't we risk a security issue if the return statement is hit?
I suggest simply raising [APIException](https://github.com/encode/django-rest-framework/blob/master/rest_framework/exceptions.py#L96-L130) if the except block is reached.

nemesisdesign (changes requested):

openwisp_users/api/mixins.py
ManishShah120: > @ManishShah120 why would the request be missing from the serializer context?

@nemesisdesign when we use nested serializer, then the `self.context` dict is `empty` when the server is started so it cannot get the `request` object from the context hence `KeyError` is raised. But once the server has started successfully then when send a request to the endpoint containing the nested serializer is sent, then the `self.context` dict contains all the required objects.

We are required to so because the server doesn't start, due to this KeyError being raised.

I tried raising `APIException` but again the server doesn't start raising this error:-
```
  File "/home/manish/Videos/openwisp_users_remote/openwisp-users/openwisp_users/api/mixins.py", line 132, in filter_fields
    raise APIException
rest_framework.exceptions.APIException: A server error occurred.
```

openwisp_users/api/mixins.py
ManishShah120: Done @atb00ker :smile:

openwisp_users/api/mixins.py
nemesisdesign: let me know how to replicate this issue and I'll test it

nemesisdesign (changes requested):
@ManishShah120 some conflicts arose after merging https://github.com/openwisp/openwisp-users/pull/258

openwisp_users/api/mixins.py
ManishShah120: @nemesisdesign To replicate this issue simply fetch the changes I have pushed to this branch https://github.com/openwisp/openwisp-firmware-upgrader/tree/openwisp-users-api-mixins ,and update openwisp-users to latest master. Then start the server.

Ps: This error is raised when we use `FilterSerializerByOrgManaged` in a serializer class which contains nested serilaizer field.

Or you can undo the changes and see the tests.

tests/testapp/urls.py
```diff
@@ -51,4 +51,5 @@
     path('template/<int:pk>/', views.template_detail, name='test_template_detail',),
     path('library/', views.library_list, name='test_library_list',),
     path('library/<int:pk>/', views.library_detail, name='test_library_detail',),
+    path('booknestedshelf/', views.book_nested_shelf, name='test_book_nested_shelf',),
```
atb00ker, line 54: Okay, for this one, why did we need another endpoint?

tests/testapp/tests/test_filter_classes.py
```diff
@@ -244,3 +245,14 @@ def test_filter_by_org_managed_with_org_field(self):
...
+
+    def test_book_nested_shelf(self):
+        operator = self._get_operator()
+        self._create_org_user(
+            user=operator, is_admin=True, organization=self._get_org('org_a')
+        )
+        token = self._obtain_auth_token(operator)
+        with self.assertNumQueries(0):
+            url = reverse('test_book_nested_shelf')
+        response = self.client.get(url, HTTP_AUTHORIZATION=f'Bearer {token}')
```
atb00ker, line 257: Wait, `self.client.get(url, HTTP_AUTHORIZATION=f'Bearer {token}')` should be inside `self.assertNumQueries(0)`, can you please explain this?

openwisp_users/api/mixins.py
```diff
@@ -131,7 +131,11 @@ def _user_attr(self):
...
 
     def filter_fields(self):
-        user = self.context['request'].user
+        try:
+            user = self.context['request'].user
+        except KeyError:
+            return
+
         if user.is_superuser:
             return
```
atb00ker, line 140: Possible improvement:
```
user = self.context.get('request', {}).get('user', None)
if !user or user.is_superuser:
   return
```

Will this work?

atb00ker (commented):
Please checkout the comments! :smile:

openwisp_users/api/mixins.py
nemesisdesign: try/except is more suited to exceptional cases

tests/testapp/tests/test_filter_classes.py
nemesisdesign: Ajay is right ,there's a mistake here

tests/testapp/urls.py
ManishShah120: This error is raised only with the serializer classes having nested serializer field. Also see:- https://github.com/openwisp/openwisp-users/pull/260#discussion_r652607748

tests/testapp/tests/test_filter_classes.py
ManishShah120: Ohh sorry, will try to rectify it ASAP.

openwisp_users/api/mixins.py
ManishShah120: > Possible improvement:
> 
> ```
> user = self.context.get('request', {}).get('user', None)
> if !user or user.is_superuser:
>    return
> ```
> 
> Will this work?

@atb00ker I tired implementing it I get this error `AttributeError: 'Request' object has no attribute 'get'`

atb00ker: Try this:
```
request = self.context['request']
user request.get('user', None)
```
If this doesn't work, I can try doing it on my machine, but I would recommend to manipulate it to create a solution like this yourself! :smile:

ManishShah120: @atb00ker Will this be a good replacement??

``` diff
diff --git a/openwisp_users/api/mixins.py b/openwisp_users/api/mixins.py
index 2bfbda8..524649b 100644
--- a/openwisp_users/api/mixins.py
+++ b/openwisp_users/api/mixins.py
@@ -131,9 +131,9 @@ class FilterSerializerByOrganization(OrgLookup):
         raise NotImplementedError()
 
     def filter_fields(self):
-        try:
+        if self.context:
             user = self.context['request'].user
-        except KeyError:
+        else:
             return
```
I tried raising an exception in the else block but, since the error is raised only when the server is started and not when it is running, so it didn't work.

atb00ker: Yes, I think that should do the trick as well! :smile:

atb00ker (dismissed):
Just the one suggested change required, rest looks good to me.

atb00ker (approved):
LGTM! :smile:

tests/testapp/urls.py
atb00ker: Okay, can you please go to this endpoint and try to create a book nested shelf?
I am not able to do that...
```
{
    "shelf": [
        "This field cannot be null."
    ]
}
```

atb00ker (changes requested):

tests/testapp/serializers.py
```diff
@@ -56,3 +56,20 @@ class LibrarySerializer(FilterSerializerByOrgManaged, ValidatedModelSerializer):
...
+class ShelfSerializerForBook(FilterSerializerByOrgManaged, ValidatedModelSerializer):
+    class Meta:
+        model = Shelf
+        fields = ['name']
+
+
+class BookWithNestedShelfSerializer(
+    FilterSerializerByOrgManaged, ValidatedModelSerializer
+):
+    shelf = ShelfSerializerForBook(read_only=True)
```
ManishShah120, line 70: Hi, @atb00ker I haven't wrote the code to support writable nested serilaizer, this is only `read_only` nested serializer. Hence I wrote the tests for get requests, only.

tests/testapp/urls.py
ManishShah120: https://github.com/openwisp/openwisp-users/pull/260/files#r659296852

tests/testapp/serializers.py
atb00ker: But we need that code for testing, we can't create an endpoint that doesn't work.

ManishShah120: Done

nemesisdesign (commented):
@ManishShah120 @atb00ker in this last commit I changed the code to simply not call filter_fields if the logic is not being called from an HTTP request: https://github.com/openwisp/openwisp-users/pull/260/commits/331f3aebceed30817873aa923fdbf54b36144fd6

Please take a look and let me know what you think.
If there's no other technical issues I am not aware of, I think this kind of approach is a lot easier for anyone to understand:

- if it's an HTTP request we do the filtering
- if it's not an HTTP request then it's just initialization and we do nothing

No doubts then arise regarding security! It's pretty clear.
A small change in the approach makes the code more readable IMHO.

ManishShah120:
> @ManishShah120 @atb00ker in this last commit I changed the code to simply not call filter_fields if the logic is not being called from an HTTP request: [331f3ae](https://github.com/openwisp/openwisp-users/commit/331f3aebceed30817873aa923fdbf54b36144fd6)
> 
> Please take a look and let me know what you think.
> If there's no other technical issues I am not aware of, I think this kind of approach is a lot easier for anyone to understand:
> 
> * if it's an HTTP request we do the filtering
> * if it's not an HTTP request then it's just initialization and we do nothing
> 
> No doubts then arise regarding security! It's pretty clear.
> A small change in the approach makes the code more readable IMHO.


Yes, @nemesisdesign this approach looks more precise, I tested it.

nemesisdesign (approved):

PR #264, opened 2021-07-01
[WIP] [api] Rest API for `openwisp-users` module #250

### LINKED ISSUE #250: [API] Create API endpoints for the `openwisp-users` module
The API endpoints should allow:-
- users to change and update permissions, groups, organization-users
- An extra endpoint to manage email addresses of users
- Endpoints to allow CRUD operations for `organizations`

ManishShah120:
Closes #250


ManishShah120:
Hi, @atb00ker can you please have a quick look, If I am proceeding in the right direction.

openwisp_users/api/urls.py
```diff
@@ -9,6 +10,16 @@ def get_api_urls(api_views=None):
...
     if api_views is None:
         api_views = views
+    urlpatterns += [
+        path('user/organization/', views.organization_list, name='organization_list',),
+        path(
+            'user/organization/<str:pk>/',
+            views.organization_detail,
+            name='organization_detail',
+        ),
+        path('user/users/', views.users_list, name='user_list',),
```
atb00ker, line 20: @nemesisdesign , hey, what do you think about the endpoints `/user/organization/` and `/user/users`?
I would personally prefer something other than that `/user/users`, but what would your advice be?

openwisp_users/api/views.py
```diff
@@ -20,4 +39,62 @@ def post(self, request, *args, **kwargs):
...
+            return User.objects.all()
+
+        if not user.is_superuser:
+            org_users = OrganizationUser.objects.filter(user=user).select_related(
+                'organization'
+            )
+            qs = User.objects.none()
+            for org_user in org_users:
+                if org_user.is_admin:
+                    qs = qs | org_user.organization.users.all().distinct()
```
atb00ker, line 83: Do we need a `distinct` here?

openwisp_users/api/urls.py
nemesisdesign: in general we keep resource name in the singluar form, there are some exceptions but not in this case.

If we wanted to prefix the module, we should rather use users/ from openwisp-users.

So it could be: /users/user/, /users/organization/, /users/group/, etc.

If we want to keep consistency with what we are doing in the other modules, this is probably the best we can do, do you have any alterantive in mind?

atb00ker: perhaps `/iam/user` or `/identity/user`; `users/user`, the same word repeated just looks odd, but it works as well!

nemesisdesign: iam and identity are two words we don't use at all

nemesisdesign (changes requested):
Good progress @ManishShah120!

The name of the module is `users`, while user is the name of the resource (we use the singular form), therefore to stay consistent with the naming scheme we are using across all modules, the prefix should be `/api/v1/users/` and the user endpoints /api/v1/users/user/`.

**Question**: how can organization owners be changed?

README.rst
```diff
@@ -331,6 +331,105 @@ the token in the ``Authorization`` header:
...
+and in the `Browsable web page <#browsable-web-interface>`_ of each point,
+here we'll provide just a list of the available endpoints,
+for further information please open the URL of the endpoint in your browser.
+
+Change User password
+^^^^^^^^^^^^^^^^^^^^
+
+.. code-block:: text
+
+    PUT /api/v1/user/users/{id}/changepassword/
```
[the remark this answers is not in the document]
ManishShah120: Hi, GET is allowed, because I am displaying the ID & Username of the user's whose password will be changed. If that is not required, should I remove it and allow PUT request only?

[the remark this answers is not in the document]
nemesisdesign: I think only PUT should be allowed here, it's the only one that it's really needed.

openwisp_users/tests/test_api/test_api.py
```diff
@@ -0,0 +1,627 @@
...
+        with self.assertNumQueries(11):
+            r = self.client.patch(path, data, content_type='application/json')
+        self.assertEqual(r.status_code, 200)
+        self.assertEqual(r.data['owner'], None)
+
+    def test_organization_delete_api(self):
+        org1 = self._create_org(name='test org 2')
+        self.assertEqual(Organization.objects.count(), 2)
+        path = reverse('users:organization_detail', args=(org1.pk,))
+        r = self.client.delete(path)
```
ManishShah120, line 131: Hi, In some of the tests, I have not asserted the number of queries, because it was changing on different test commands, the tests were passing on some and failing on some.

openwisp_users/api/serializers.py
```diff
@@ -0,0 +1,402 @@
...
+            'slug',
+            'description',
+            'email',
+            'url',
+            'owner',
+            'created',
+            'modified',
+        )
+
+    def update(self, instance, validated_data):
```
atb00ker, line 79: Things look orthodox, except this function would need some testing.
@ManishShah120 I hope you have already done functional testing, as done before in other PRs! 😄

openwisp_users/api/urls.py
```diff
@@ -9,6 +10,24 @@ def get_api_urls(api_views=None):
...
         api_views = views
+    urlpatterns += [
+        path('users/organization/', views.organization_list, name='organization_list',),
+        path(
+            'users/organization/<str:pk>/',
+            views.organization_detail,
+            name='organization_detail',
+        ),
+        path('users/user/', views.user_list, name='user_list',),
+        path('users/user/<str:pk>/', views.user_detail, name='user_detail'),
```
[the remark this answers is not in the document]
atb00ker: We are showing the hashed password, do we want to send it back like this @nemesisdesign @ManishShah120 ? I would assume we wouldn't share this in API even though the UI shows it; I don't know in which case sharing the password would be helpful! 
ss4

openwisp_users/api/urls.py
```diff
@@ -9,6 +10,24 @@ def get_api_urls(api_views=None):
...
+        path(
```
atb00ker, line 22: When I submit form without any information, it returns that I have entered wrong information.
This might be confusing when the consumer of the API has a typo or something, we should return, "please enter old and new  passwords".

openwisp_users/api/urls.py
```diff
@@ -9,6 +10,24 @@ def get_api_urls(api_views=None):
...
+            'users/user/<str:pk>/password/',
+            views.change_password,
+            name='change_password',
+        ),
+        path('users/user/<str:pk>/email/', views.email_update, name='email_update',),
+        path('users/group/', views.group_list, name='group_list'),
+        path('users/group/<str:pk>/', views.group_detail, name='group_detail'),
```
atb00ker, line 29: Shouldn't this be `users/group/<int:pk>/`, group ids are in integers.

atb00ker (commented):
Checked some endpoints:

okraits (commented):
I also reviewed and tested this. I can only confirm Ajay's findings.

openwisp_users/api/urls.py
ManishShah120: Fixed, :+1:

openwisp_users/api/urls.py
ManishShah120: Fixed, :+1:

openwisp_users/api/urls.py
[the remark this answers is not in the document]
ManishShah120: > Create a new user with organizations value.
> Open the user with UUID in API UI, you'll notice while the value is present in json, form is empty:

Yes, @atb00ker, the form is present to add or remove org user(if the sent data is same then it will get removed else, the changes will be accomplished, this field would have been populated if there would be m2m relationship, for example what would happen if the user is OrgUser of more than one Org?? :smile:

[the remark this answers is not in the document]
ManishShah120: Org1admin cannot see organization user's section!
ss4

While an org_users connection exists :

            "organization_users": [
                {
                    "is_admin": false,
                    "organization": "091dbab0-4188-4e83-a2e3-40853c9189f4"
                }
            ]
Yes, this is intentinal, the Org1admin, can only see the the org_users, present but cannot edit anything, so I removed this field for it.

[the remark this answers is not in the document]
ManishShah120: > Create an org owner and try to downgrade it, not handled gracefully.

@atb00ker This is already been handled by the `full_clean` method, so will it be not redundant to repeat this, in usecase/production, it will just return the error. I guesss??

openwisp_users/api/views.py
ManishShah120: I have implemented this logic from here https://github.com/openwisp/openwisp-users/blob/master/openwisp_users/multitenancy.py#L99-L119

openwisp_users/api/urls.py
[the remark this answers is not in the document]
atb00ker: > @atb00ker This is already been handled by the full_clean method, so will it be not redundant to repeat this, in usecase/production, it will just return the error in normal way. I guess this is what we did in the config App!!!

I belive the response should be graceful when possible. Unknown cases can happily return 500, but a known case should give error message and explain the message! 😄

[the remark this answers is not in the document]
atb00ker: > Yes, this is intentinal, the Org1admin, can only see the the org_users, present but cannot edit anything, so I removed this field for it.

If the user is not allowed to edit something, we should have it as readonly for consistency, what do you think? (Also, this can be lower priority)

okraits (commented):
If I access the live API documentation at `/api/v1/docs/` I get the following exceptions:

> view's EmailUpdateView raised exception during schema generation; use `getattr(self, 'swagger_fake_view', False)` to detect and short-circuit this
> AssertionError: Expected view EmailUpdateView to be called with a URL keyword argument named "pk". Fix your URL conf, or set the `.lookup_field` attribute on the view correctly.
> view's EmailUpdateView raised exception during schema generation; use `getattr(self, 'swagger_fake_view', False)` to detect and short-circuit this
> view's ChangePasswordView raised exception during schema generation; use `getattr(self, 'swagger_fake_view', False)` to detect and short-circuit this
>   File "/home/okraitschy/work/gsoc/openwisp-users/openwisp_users/api/views.py", line 177, in get_object
>     'id': self.kwargs['pk'],
> KeyError: 'pk'

ManishShah120:
> If I access the live API documentation at `/api/v1/docs/` I get the following exceptions:
> 
> > view's EmailUpdateView raised exception during schema generation; use `getattr(self, 'swagger_fake_view', False)` to detect and short-circuit this
> > AssertionError: Expected view EmailUpdateView to be called with a URL keyword argument named "pk". Fix your URL conf, or set the `.lookup_field` attribute on the view correctly.
> > view's EmailUpdateView raised exception during schema generation; use `getattr(self, 'swagger_fake_view', False)` to detect and short-circuit this
> > view's ChangePasswordView raised exception during schema generation; use `getattr(self, 'swagger_fake_view', False)` to detect and short-circuit this
> > File "/home/okraitschy/work/gsoc/openwisp-users/openwisp_users/api/views.py", line 177, in get_object
> > 'id': self.kwargs['pk'],
> > KeyError: 'pk'

Ok, trying to recreate the issue.

nemesisdesign (changes requested):
> If I access the live API documentation at `/api/v1/docs/` I get the following exceptions:
> 
> > view's EmailUpdateView raised exception during schema generation; use `getattr(self, 'swagger_fake_view', False)` to detect and short-circuit this
> > AssertionError: Expected view EmailUpdateView to be called with a URL keyword argument named "pk". Fix your URL conf, or set the `.lookup_field` attribute on the view correctly.
> > view's EmailUpdateView raised exception during schema generation; use `getattr(self, 'swagger_fake_view', False)` to detect and short-circuit this
> > view's ChangePasswordView raised exception during schema generation; use `getattr(self, 'swagger_fake_view', False)` to detect and short-circuit this
> > File "/home/okraitschy/work/gsoc/openwisp-users/openwisp_users/api/views.py", line 177, in get_object
> > 'id': self.kwargs['pk'],
> > KeyError: 'pk'

When I open `/api/v1/docs/` I see a lot of error output in the dev server, although the page opens.

### Email endpoint

This endpoint should allow to manage the email addresses of the user, 
which can be multiple addresses, but here I see only one address can be edited,
this is not compliant with what was requested, it was requested to do it as the admin
and is not working as in the admin.

The response of this endpoint should be a list of objects, not an object.

Each email shall be editable:

/api/v1/users/user/{pk}/email/ ListCreateApiView
/api/v1/users/user/{pk}/email/{email_pk}/ RetrieveUpdateDestroyView

### Password endpoint notes

This is the current endpoint:

```
{
    "old_password": "",
    "new_password": ""
}
```

I would change ``old_password`` to ``current_password``.

I think we need a confirm password field like we have everywhere we deal with passwords (also in OpenWISP RADIUS).

README.rst
```diff
@@ -331,6 +331,182 @@ the token in the ``Authorization`` header:
...
+and in the `Browsable web page <#browsable-web-interface>`_ of each point,
+here we'll provide just a list of the available endpoints,
+for further information please open the URL of the endpoint in your browser.
+
+Change User password
+^^^^^^^^^^^^^^^^^^^^
+
+.. code-block:: text
+
+    PUT /api/v1/users/user/{id}/changepassword/
```
nemesisdesign, line 347: this looks outdated, please make sure the docs are up to date with the code. Every time you change the code you have to update the docs.

nemesisdesign (changes requested):

nemesisdesign (changes requested):
I did a quick test on the email endpoint, thanks for implementing the changes!
We should remove the possibility of flagging the user as verified because otherwise anyone would be able to flag their email as verified.
Verification is done by other means (we may have to add another API endpoint just for this later on because the verification link opens a page in django, but we can leave that for a later phase).

ManishShah120:
> I did a quick test on the email endpoint, thanks for implementing the changes!
> We should remove the possibility of flagging the user as verified because otherwise anyone would be able to flag their email as verified.
> Verification is done by other means (we may have to add another API endpoint just for this later on because the verification link opens a page in django, but we can leave that for a later phase).

Implemented it.

nemesisdesign (commented):
> > I did a quick test on the email endpoint, thanks for implementing the changes!
> > We should remove the possibility of flagging the user as verified because otherwise anyone would be able to flag their email as verified.
> > Verification is done by other means (we may have to add another API endpoint just for this later on because the verification link opens a page in django, but we can leave that for a later phase).
> 
> Implemented it.

These problems are fixed, thanks :+1:

openwisp_users/api/serializers.py
```diff
@@ -0,0 +1,434 @@
...
+    )
+
+    def validate_confirm_password(self, value):
+        if self.initial_data.get('new_password') != self.initial_data.get(
+            'confirm_password'
+        ):
+            raise serializers.ValidationError(
+                _('New password and Confirm password do not match.')
+            )
+        return value
```
codesankalp, line 397: @ManishShah120 I think there must be a check that `new_password` or `confirm_password` can't be the same as `current_password`. What do you think?

openwisp_users/api/views.py
atb00ker: If the code is trying to do something so similar at 2 places, why is it not shared? 
i.e creating a mixin to share the logic appropriately! :smile:

openwisp_users/api/serializers.py
nemesisdesign: > @ManishShah120 I think there must be a check that `new_password` or `confirm_password` can't be the same as `current_password`. What do you think?

Agreed :+1:·

ManishShah120: Yes, will add this check ASAP, thnx @codesankalp for pointing this out. :+1:

openwisp_users/api/views.py
ManishShah120: Sure, @atb00ker , I will  try to move this logic to a mixin. ASAP. :smile:

openwisp_users/api/serializers.py
ManishShah120: Done adding the check for same old and new password. :smile:

PR #273, opened 2021-08-14
[fix] Fixed breaking to tests due to new filter #272

### LINKED ISSUE #272: [bug] New filters breaking tests
https://github.com/openwisp/openwisp-users/blob/08c24f303cff77baa86dc3ddf80fa4099afa77e1/openwisp_users/tests/test_admin.py#L1721-L1750

The above test has become flaky after inclusion of new filters.
The underlying HTML elements differs when number of filters changes (>4), this makes the test fail in different modules. 

Example:
The [build is falling](https://github.com/openwisp/openwisp-radius/runs/3323840897?check_suite_focus=true) because of this in openwisp-radius because it shows more than 4 filters even for non superusers.

niteshsinha17:
closes #272

### DIFF
--- a/openwisp_users/tests/test_admin.py
+++ b/openwisp_users/tests/test_admin.py
@@ -1728,16 +1728,16 @@ class TestMultitenantAdmin(TestMultitenantAdminMixin, TestOrganizationMixin, Tes
             self.client.force_login(user)
             response = self.client.get(url)
             for org in Organization.objects.all():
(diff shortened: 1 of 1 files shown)


niteshsinha17:
@nemesisdesign @pandafy After merging of this https://github.com/openwisp/openwisp-utils/pull/242 tests will pass.

pandafy (approved):

nemesisdesign (approved):
Thanks!

PR #294, opened 2021-11-03
[ux/i18n] Improved serializer strings and added italian translations

### LINKED ISSUE #293: [ux/i18n] User API serializer strings need to be improved and translated
The wording used in https://github.com/openwisp/openwisp-users/blob/master/openwisp_users/api/serializers.py for some API error messages are unusual and need to be improved to be more consistent with what is used in the rest of the application and in django.

The messages also need to be translated (we can use google translate for that) because some of those messages will be shown to the users in https://github.com/openwisp/openwisp-wifi-login-pages.

codesankalp:
Closes #293


openwisp_users/api/serializers.py
```diff
@@ -392,7 +392,7 @@ def validate_confirm_password(self, value):
             'confirm_password'
         ):
             raise serializers.ValidationError(
-                _('New password and Confirm password do not match.')
+                _('The two password fields didn’t match.')
```
nemesisdesign, line 395: `didn’t` seems to contain a slant quote here, I suggest replacing it with a normal quote for consistency

openwisp_users/api/locale/it/LC_MESSAGES/django.po
```diff
@@ -0,0 +1,81 @@
...
+msgstr ""
+"L'utente non è un proprietario dell'organizzazione a cui la risorsa richiesta "
+"appartiene."
+
+#: serializers.py:227
+msgid "This field may not be blank."
+msgstr "Questo campo non può essere vuoto."
+
+#: serializers.py:395
+msgid "The two password fields didn’t match."
```
nemesisdesign, line 64: same here

openwisp_users/api/locale/it/LC_MESSAGES/django.po
```diff
@@ -0,0 +1,81 @@
...
+msgid ""
+"User is not a owner of the organization to which the requested resource "
+"belongs."
+msgstr ""
+"L'utente non è un proprietario dell'organizzazione a cui la risorsa richiesta "
+"appartiene."
+
+#: serializers.py:227
+msgid "This field may not be blank."
+msgstr "Questo campo non può essere vuoto."
```
nemesisdesign, line 61: It's possible that this msgid is already present in the translations included in Django, please double check

openwisp_users/api/locale/it/LC_MESSAGES/django.po
nemesisdesign: Can you also double check if a similar msgid is already used in Django and in that case switch to the exact form used in Django? That would also allow to get rid of the translations.

openwisp_users/api/locale/it/LC_MESSAGES/django.po
```diff
@@ -0,0 +1,81 @@
...
+"Report-Msgid-Bugs-To: \n"
+"POT-Creation-Date: 2021-11-03 16:47+0530\n"
+"PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
+"Last-Translator: FULL NAME <EMAIL@ADDRESS>\n"
+"Language-Team: LANGUAGE <LL@li.org>\n"
+"Language: \n"
+"MIME-Version: 1.0\n"
+"Content-Type: text/plain; charset=UTF-8\n"
+"Content-Transfer-Encoding: 8bit\n"
+"Plural-Forms: nplurals=2; plural=(n != 1);\n"
```
nemesisdesign, line 19: Let's do it similar to https://github.com/openwisp/openwisp-users/blob/master/openwisp_users/accounts/locale/it/LC_MESSAGES/django.po#L8-L13 for consistency.

nemesisdesign (changes requested):
Looks good, see below for a few requests for improvment.

Can you please add translations for the other locales already supported?
See https://github.com/openwisp/openwisp-users/tree/master/openwisp_users/accounts/locale.

openwisp_users/api/locale/it/LC_MESSAGES/django.po
codesankalp: Yes, it's registered. https://github.com/django/django/blob/4816dc942860caf076c7c85ea9dbfa8bfab212ff/django/contrib/auth/forms.py#L397-L399

openwisp_users/api/locale/it/LC_MESSAGES/django.po
codesankalp: Replacing it with `This field cannot be blank.`

openwisp_users/api/serializers.py
codesankalp: This slant quote is also present in Django https://github.com/django/django/blob/4816dc942860caf076c7c85ea9dbfa8bfab212ff/django/contrib/auth/forms.py#L397-L399

nemesisdesign (commented):
Looks good, I'll do some improvements and more testing asap. Thanks!

nemesisdesign (changes requested):
The translations have been added to a wrong location: `openwisp_users/api/locale`, it should be `openwisp_users/locale`, because the app is `openwisp_users/`, not `openwisp_users/api/`, translations are loaded at django app level, `openwisp_users` and `openwisp_users.accounts` are two django apps, `openwisp_users.api` is not a django app and therefore translations do not get loaded.

I am trying with a browser which has Italian as preferred language, testing through the swagger API docs doesn't return italian translations for these messages, although italian translations are returned for messages which are included in django and other apps.

While testing this I also figured out that some translations are not needed because already included in django, see below for more info.

nemesisdesign (changes requested):
can you please fix the commit history in this branch so that `[chores] Formatted code with new black version` is not included? It's already included in master.

nemesisdesign (commented):
The translations now work :+1:

nemesisdesign (approved):

PR #295, opened 2021-11-03
[change] Use html template in email confirmation #285

### LINKED ISSUE #285: [feature] Use HTML template to send email confirmation
Depends on https://github.com/openwisp/openwisp-utils/issues/259.

According to [django-allauth docs](https://django-allauth.readthedocs.io/en/latest/advanced.html#sending-email), we can send the email confirmation message using HTML.

We should do so and use the same HTML design which will be defined in openwisp-utils, so that communications via email to users will have a consistent look.

devkapilbansal:
Closes #285


openwisp_users/accounts/templates/account/email/email_confirmation_message.html
```diff
@@ -0,0 +1,11 @@
+{% extends "account/email/base_message.html" %}
+{% load account %}
+
+{% block mail_body %}
+{{ block.super }}
+{% user_display user as user_display %}
+{% blocktrans with site_name=current_site.name site_domain=current_site.name %}
+You're receiving this e-mail because user {{ user_display }} registered this e-mail address to create an account on {{ site_domain|safe }}.
```
nemesisdesign, line 8: Let's put the username and domain name in bold.

openwisp_users/accounts/templates/account/email/email_confirmation_message.txt
```diff
@@ -1,7 +0,0 @@
-{% extends "account/email/base_message.txt" %}
-{% load account %}
-{% load i18n %}
-
-{% block content %}{% autoescape off %}{% user_display user as user_display %}{% blocktrans with site_name=current_site.name site_domain=current_site.name %}You're receiving this e-mail because user {{ user_display }} registered this e-mail address to create an account on {{ site_domain }}.
```
nemesisdesign, line 5: To confirm this is correct, click on the button below.

And then below let's add the call to action button like "Confirm"

openwisp_users/accounts/templates/account/email/base_message.html
```diff
@@ -0,0 +1,19 @@
...
+{% block mail_body %}
+{{ block.super }}
+{% user_display user as user_display %}
+{% blocktrans with site_name=current_site.name site_domain=current_site.name %}
+You're receiving this e-mail because user {{ user_display }} registered this e-mail address to create an account on {{ site_domain|safe }}.
+To confirm this is correct, go to {{ activate_url }}
+{% endblocktrans %}
+{% blocktrans with site_name=current_site.name site_domain=current_site.name %}Thank you for using {{ site_name|safe }}!
+{{ site_domain }}{% endblocktrans %}
+{% endblock %}
```
nemesisdesign, line 19: doesn't look as a base template

nemesisdesign (commented):
Let's leverage the power of HTML and show the call to action button instead of the raw URL.

devkapilbansal:
Looks like `all_auth` is not using the html templates that I created. :thinking:  Will look into it again soon

nemesisdesign (changes requested):
we need a test which ensures that when these emails are sent, the HTML templates are used

openwisp_users/accounts/templates/account/email/base_message.html
```diff
@@ -0,0 +1,23 @@
+{% extends "openwisp_utils/email_template.html" %}
+{% load i18n %}
+{% block body %}
+  <div class="container">
+    <div class="box">
+      {% block mail_header %}
+      <div class="logo-container">
+        <img src="https://raw.githubusercontent.com/openwisp/openwisp-utils/issues/259-email-template/openwisp_utils/static/openwisp-utils/images/openwisp-logo.png" alt="Logo" id="logo" class="logo">
```
nemesisdesign, line 8: why are you redefining this? This defeats the purpose of the reusable template, if we configure a different logo, the configuration will have no effect because of this.

openwisp_users/accounts/templates/account/email/base_message.html
```diff
@@ -0,0 +1,23 @@
...
+      </div>
+      <div class="subject">
+        {% blocktrans with site_name=current_site.name %}Welcome to <b>{{ site_name }}!</b>{% endblocktrans %}
+      </div>
+      {% endblock mail_header %}
+      {% block mail_body %}
```
nemesisdesign, line 14: why is this also redefined?

openwisp_users/accounts/templates/account/email/email_confirmation_signup_message.html
```diff
@@ -0,0 +1,15 @@
...
+
+{% block mail_body %}
+{% user_display user as user_display %}
+{% blocktrans with site_name=current_site.name site_domain=current_site.name %}
+You're receiving this e-mail because user <b>{{ user_display }}</b> registered this e-mail address to create an account on <b>{{ site_domain }}</b>.
+To confirm this is correct, click on the button below.
+{% endblocktrans %}
+<div class="call-to-action">
+    <a href="{{ activate_url }}" class="btn">
+      Confirm</a>
```
nemesisdesign, line 13: Don't forget to flag this as translatable.

openwisp_users/accounts/templates/account/email/email_confirmation_signup_message.html
```diff
@@ -0,0 +1,15 @@
+{% extends "account/email/base_message.html" %}
+{% load account %}
+{% load i18n %}
+
+{% block mail_body %}
+{% user_display user as user_display %}
+{% blocktrans with site_name=current_site.name site_domain=current_site.name %}
+You're receiving this e-mail because user <b>{{ user_display }}</b> registered this e-mail address to create an account on <b>{{ site_domain }}</b>.
+To confirm this is correct, click on the button below.
```
nemesisdesign, line 9: please click on the "confirm" button below.

nemesisdesign (changes requested):
I think we're still not quite there yet, quite a few elements of the HTML theme are redefined in this code, please look at https://github.com/openwisp/openwisp-radius/pull/337/files#diff-96f72a896798cf0b0f8773420125ae362b17c77a4e4bbca656a62ce8d888f225R1-R35 to see how Sankalp did it there.

nemesisdesign (changes requested):
Regarding the adapter, see https://github.com/pennersr/django-allauth/blob/master/allauth/account/adapter.py#L134-L136
It should be possible to subclass that adapter and modify that method (and others if needed) to call the new send_email function shipped in openwisp-utils

then we can make the new adapter the default with a trick in openwisp_utils.admin_theme (so that we avoid having to update all the deployment stuff), eg:
https://github.com/openwisp/openwisp-users/blob/0bc275232c6946017968a1a06db978eb5fd361cb/openwisp_users/apps.py#L74-L88

openwisp_users/apps.py
```diff
@@ -87,6 +87,9 @@ def set_default_settings(self):
             }
             setattr(settings, 'SWAGGER_SETTINGS', SWAGGER_SETTINGS)
 
+        ACCOUNT_ADAPTER = "openwisp_users.accounts.adapter.EmailAdapter"
```
nemesisdesign, line 90: let's do as `LOGIN_URL` and `LOGOUT_URL`, set this only if not present in settings (we allow users to customize it if needed).

Please ensure quoting style is consistent (single quotes).

openwisp_users/accounts/adapter.py
```diff
@@ -0,0 +1,16 @@
+from allauth.account.adapter import DefaultAccountAdapter
+from django.utils.translation import gettext as _
+from openwisp_utils.admin_theme import settings as utils_settings
+
+
+class EmailAdapter(DefaultAccountAdapter):
+    def send_mail(self, template_prefix, email, context):
+        context["logo_url"] = utils_settings.OPENWISP_EMAIL_LOGO
+        site_name = context["current_site"].name
+        context["subject"] = _("Welcome to %(site_name)s") % {"site_name": site_name}
```
nemesisdesign, line 10: please ensure quoting style is consistent with the rest of the codebase (single quotes).

openwisp_users/accounts/templates/account/email/base_message.html
```diff
@@ -0,0 +1 @@
+{% extends "openwisp_utils/email_template.html" %}
```
nemesisdesign, line 1: here we could avoid extending from this template and either leave this empty or remove it

openwisp_users/accounts/templates/account/email/email_confirmation_message.txt
```diff
@@ -1,7 +0,0 @@
...
-
-To confirm this is correct, go to {{ activate_url }}{% endblocktrans %}{% endautoescape %}{% endblock %}
```
nemesisdesign, line 7: This template could be left so we can send the text version.

openwisp_users/accounts/templates/account/email/email_confirmation_signup_message.html
```diff
@@ -0,0 +1,18 @@
...
+{% blocktrans with site_name=current_site.name site_domain=current_site.name %}
+You're receiving this e-mail because user <b>{{ user_display }}</b> registered this e-mail address to create an account on <b>{{ site_domain }}</b>.
+To confirm this is correct, please click on the button below.
+{% endblocktrans %}
+</div>
+{% endblock mail_body %}
+{% block mail_footer %}
+{% blocktrans with site_name=current_site.name %}
+<div class="msg">Thank you for using <b>{{site_name}}</b></div>
+{% endblocktrans %}
```
nemesisdesign, line 17: Here we would include only the HTML which would be included in the message

openwisp_users/accounts/adapter.py
```diff
@@ -0,0 +1,16 @@
...
+        context["message"] = _("Thank you for using <b>%(site_name)s</b>") % {
+            "site_name": site_name
+        }
+        context["call_to_action_url"] = context["activate_url"]
+        context["call_to_action_text"] = _("Confirm")
+        super().send_mail(template_prefix, email, context)
```
nemesisdesign, line 16: You know what bugs me about this solution?
If we change our `send_email` and/or our templates in openwisp-utils, we may have to come back to here and update this too, which means double work for us, which is a very unpleasant solution to be in, because we'll be uncentivized to make improvements and we have to do our best to avoid this and I think it's avoidable by structuring the code differently.

I think we could try a different approach which will allow us to make improvements in the email theme feature of openwisp-utils without having to come back here to update this code.

Let's look at the `send_mail()` method of the current default adapter of allauth:

https://github.com/pennersr/django-allauth/blob/2fa7038a53e77dad504624584d2a533d08a3e7d1/allauth/account/adapter.py#L99-L136

If you take a close look, you will notice that it basically does almost the same operations we now do in openwisp-utils, except that our function is reusable across different modules, the logic of allauth only applies to allauth.

I believe we should simply rewrite the `send_mail()` method of our adapter to use the `send_email` function of openwisp-utils.

As allauth does, we should look for both the `html` and `txt` template files and render both, so we can send both text and html versions.

The difference is that in the HTML template we won't extend the base HTML email theme of openwisp-utils, but we'll generate only the message part. See my following comments for more info.

nemesisdesign (changes requested):
Good progress Kapil, now is starting to look better.

Please see my comments below.

openwisp_users/accounts/templates/account/email/email_confirmation_signup_message.html
```diff
@@ -1,4 +1,4 @@
-{% extends "account/email/base_message.html" %}
+{% extends "openwisp_utils/email_template.html" %}
```
nemesisdesign, line 1: I think we should not extend from `openwisp_utils/email_template.html` if we intend to pass this as the body argument to `send_email`

nemesisdesign (changes requested):
I found an issue which doesn't happen on master.

It seems this patch is impacting `http://localhost:8000/accounts/password/reset/` (the password rset feature we have available from the admin login form) negatively, if I try to use it in the dev env of this project I get an exception:

```
TemplateDoesNotExist at /accounts/password/reset/
account/email/password_reset_key_message.html
```

Can you please add a failing test for this and fix it?

nemesisdesign (commented):
PS: the rest works also in a prod-like environment :+1:!

devkapilbansal:
> I found an issue which doesn't happen on master.
> 
> It seems this patch is impacting `http://localhost:8000/accounts/password/reset/` (the password rset feature we have available from the admin login form) negatively, if I try to use it in the dev env of this project I get an exception:
> 
> ```
> TemplateDoesNotExist at /accounts/password/reset/
> account/email/password_reset_key_message.html
> ```
> 
> Can you please add a failing test for this and fix it?

Hi,
I tried it locally and the page is available :thinking: Is it possible that this comes when used with any other repository

Password Reset

nemesisdesign (commented):
I can also see the page, but when it's submitted I get error, I tried also in openwips-users test env to make sure it was not an interaction with other modules.

openwisp_users/accounts/adapter.py
```diff
@@ -0,0 +1,32 @@
...
+                content[ext] = render_to_string(
+                    template_name, context, self.request
+                ).strip()
+            except TemplateDoesNotExist as e:
+                errors[ext] = e
+            text = content.get('txt', '')
+            html = content.get('html', '')
+            # both templates fail to load, raise the exception
+            if len(errors.keys()) >= 2:
+                raise errors['txt'] from errors['html']
```
nemesisdesign, line 31: I think we need to add a test for this case to make sure it keeps working over time, eg:

```python
with self.assertRaises(TemplateException):
    EmailAdapter(....).send_mail('bogus_prefix', 'test@test.com', {})
```

nemesisdesign (changes requested):
I removed empty translations, merged the latest master and simplified send_email() (https://github.com/openwisp/openwisp-users/pull/295/commits/610d292ecea68e0b9ab09000b1020185c9275073), please look at how I simplified that method, it should be instructive for future work.

See also my comment below.

nemesisdesign (commented):
Looks good, waiting until openwisp-utils is merged so we can update the dependency and merge this too.

nemesisdesign (approved):

PR #296, opened 2021-11-10
[chores] Formatted code with new black version

pandafy:


devkapilbansal (approved):

devkapilbansal (commented):
@pandafy
Shouldn't we change the version in `requirements` file instead of `requirements-test` if both need `openwisp-utils'

nemesisdesign:
@devkapilbansal main requirements already point to utils dev: https://github.com/openwisp/openwisp-users/blob/master/requirements.txt#L6-L7

nemesisdesign (approved):

devkapilbansal:
> @devkapilbansal main requirements already point to utils dev: https://github.com/openwisp/openwisp-users/blob/master/requirements.txt#L6-L7

I mean, we can leave it unpinned too 🤔

pandafy:
> > @devkapilbansal main requirements already point to utils dev: https://github.com/openwisp/openwisp-users/blob/master/requirements.txt#L6-L7
> 
> I mean, we can leave it unpinned too thinking

I don't understand your comment completely, but unpinning it in requirement.txt will affect installation of modules, e.g. when modules are installed by dev branch of ansible-openwisp2. 

requirements-test.txt is only used for development.

PR #297, opened 2021-11-15
[feature] Add user preferred language field #261

### LINKED ISSUE #261: [feature] Add user preferred language field
We need to store an additional field to store the preferred language of the user (string, with choices defaulting to `settings.LANGUAGES`, defaulting to a callable which uses the default locale set in django settings).

This is needed because in different modules we send email to users and in some end-user applications we need to translate the content for the end users, and when sending email we don't have HTTP requests which tell us which languages are accepted by the user so we need to have this info stored in the DB.

Something like the following should do:

```python
language = models.CharField(
    max_length=10,
    choices=settings.LANGUAGES,
    default=settings.LANGUAGE_CODE
)
```

Let's ensure the migration created points the choices directly to `settings.LANGUAGES` without duplicating the whole list/tuple, otherwise it will be painful to maintain.

codesankalp:
Closes #261


openwisp_users/base/models.py
```diff
@@ -59,6 +60,11 @@ class AbstractUser(BaseUser):
     notes = models.TextField(
         _('notes'), help_text=_('notes for internal usage'), blank=True
     )
+    language = models.CharField(
+        max_length=10,
```
nemesisdesign, line 64: Let's shorten this to 8.

7 seems the maximum used by the languages configured in Django:
https://github.com/django/django/blob/main/django/conf/global_settings.py#L56-L153

We can keep 1 spare character in case some crazy language comes out but there shouldn't be other languages which have codes longer than that.

nemesisdesign (changes requested):
Looks good, see my request below.

nemesisdesign (commented):
Looks good, I need to do some manual testing.

nemesisdesign (approved):

PR #298, opened 2021-11-23
[deps] Added support for django4.0rc1

### LINKED ISSUE #292: [deps] Add support for django 4.0.x
Let's make sure this module can run with Django 4.0.

We should probably address https://github.com/openwisp/openwisp-users/issues/178.

codesankalp:
Closes #292


codesankalp:
I have to import `OrganizationInvitation` model as abstract model due to these checks:

image

openwisp_users/models.py
```diff
@@ -35,6 +36,12 @@ class Meta(AbstractOrganizationOwner.Meta):
         swapper.swappable_setting('openwisp_users', 'OrganizationOwner')
 
 
+# only needed for django-organizations~=2.x
+class OrganizationInvitation(AbstractOrganizationInvitation):
+    class Meta(AbstractOrganizationInvitation.Meta):
+        abstract = True
```
nemesisdesign, line 42: I don't think this should be abstract, I would allow this to be swappable since the start and update the README:
https://github.com/openwisp/openwisp-users#8-add-swapper-configurations

We should specify that right now we don't use this model in OpenWISP but users are free to implement custom flows with it if needed.

Some users may be interested in implementing this.

What do you think?

tests/openwisp2/sample_users/models.py
```diff
@@ -57,6 +58,12 @@ class OrganizationOwner(DetailsModel, BaseOrganizationOwner, AbstractOrganizatio
     pass
 
 
+# only needed for django-organizations~=2.x
+class OrganizationInvitation(AbstractOrganizationInvitation):
+    class Meta(AbstractOrganizationInvitation.Meta):
+        abstract = True
```
nemesisdesign, line 64: I don't think this should be abstract

nemesisdesign (changes requested):
Nice!

nemesisdesign (approved):
Thanks :+1:

PR #306, opened 2022-01-31
[fix] Error handling where user is none #262

### LINKED ISSUE #262: [bug] Add error handling for case where user,organization_user is None
While loading data from a fixture(e.g. using loaddata), `user.organization_user.user` raises `ObjectDoesNotExists` exception probably because `OrganizationUser` table is empty at that point. This makes it impossible to import data from fixtures without changing the code.

https://github.com/openwisp/openwisp-users/blob/a1e44d4a3e15f3745181473843844b9450908f4e/openwisp_users/apps.py#L85-L92

**Possible Solution**

Wrap `user.organization_user.user` in try-except block.

Aryamanz29:
`user.organization_user.user` raises `ObjectDoesNotExists` while loading data from fixture,  Added `try-except` block to avoid that.
 
Fixes #262

### DIFF
--- a/openwisp_users/apps.py
+++ b/openwisp_users/apps.py
@@ -3,7 +3,7 @@ import logging
 from django.apps import AppConfig
 from django.conf import settings
 from django.core.cache import cache
-from django.core.exceptions import ValidationError
+from django.core.exceptions import ObjectDoesNotExist, ValidationError
 from django.db import IntegrityError, transaction
 from django.db.models.signals import post_delete, post_save
 from django.utils.translation import gettext_lazy as _
@@ -119,7 +119,10 @@ class OpenwispUsersConfig(AppConfig):
         if hasattr(instance, 'user'):
             user = instance.user
         else:
-            user = instance.organization_user.user
+            try:
+                user = instance.organization_user.user
+            except ObjectDoesNotExist:
+                return
         cache_key = 'user_{}_organizations'.format(user.pk)
         cache.delete(cache_key)
         # forces caching


Aryamanz29:
Work in progress

codesankalp (approved):

Aryamanz29:
While writing a test for this issue, I was able to reproduce `ObjectDoesNotExist` exception

```bash
    return qs.get(self.field.get_reverse_related_filter(instance))
  File "/home/user/Desktop/openwisp/o-users/env/lib/python3.8/site-packages/django/db/models/query.py", line 439, in get
    raise self.model.DoesNotExist(
openwisp_users.models.DoesNotExist: Problem installing fixture '/home/user/Desktop/openwisp/o-users/openwisp-users/tests/fixtures/OrganizationOwner.json': OrganizationUser matching query does not exist.
```

but after adding my patch
```python
    def update_organizations_dict(cls, instance, **kwargs):
        if hasattr(instance, 'user'):
            user = instance.user
        else:
            #user = instance.organization_user.user
            try:
                user = instance.organization_user.user
            except ObjectDoesNotExist:
                return
```
Another exception was raised (`IntegrityError`), See below :point_down: , It happens because it looks for `"organization_user": "52b56376-0160-4aca-8ffc-2d5d669b2e0d"` foreign key (Define in `OrganizationOwner.json`), But `OrganizationUser` table at that moment was empty.

```python
 raise IntegrityError(
django.db.utils.IntegrityError: Problem installing fixtures: The row in table 'openwisp_users_organizationowner' with primary key 'c51b995103a341f4a1c5e2402ebdef95' has an invalid foreign key: openwisp_users_organizationowner.organization_user_id contains a value '52b5637601604aca8ffc2d5d669b2e0d' that does not have a corresponding value in openwisp_users_organizationuser.id.
```
- `OrganizationOwner.json`
```json
[
    {
        "model": "openwisp_users.organizationowner",
        "pk": "c51b9951-03a3-41f4-a1c5-e2402ebdef95",
        "fields": {
            "created": "2022-02-16T09:24:37.256Z",
            "modified": "2022-02-16T09:24:37.259Z",
            "organization_user": "52b56376-0160-4aca-8ffc-2d5d669b2e0d",
            "organization": "3f344a8b-9248-4af2-a44f-0e361b61c1a9"
        }
    }
]
```
I don't think we ever wanted to load `OrganizationOwner` from fixtures without having `OrganizationUser`.

Aryamanz29:
Also, I was able to load other fixtures like `OrganizationUser.json`, `User.json`, `Organization.json` without any exceptions

nemesisdesign (commented):
Surely it's not gonna be possible to insert an OrganizationOwner if there's no related OrganizationUser because it will return error. Are we trying to fix an unfixable issue @pandafy or can you narrow down what the bug is exactly so we can write an useful test?

pandafy:
I tried replicating the issue with the database dump, but it didn't occur this time.

> Surely it's not gonna be possible to insert an OrganizationOwner if there's no related OrganizationUser because it will return error.

You are right @nemesisdesign. I think because of this error that I was facing, the execution of code was not reaching to the part that raises `IntegrityError`. 

I am closing this PR and the related issue. I will re-open them, if I encountered this bug again. 

Thanks to everyone involved.

PR #310, opened 2022-03-01
[tests] Update TestMultitenantAdminMixin to use default operator group #106

### LINKED ISSUE #106: [tests] Update TestMultitenantAdminMixin to use default operator group

codesankalp:
Closes #106

### DIFF
--- a/openwisp_users/tests/utils.py
+++ b/openwisp_users/tests/utils.py
@@ -9,6 +9,7 @@ from swapper import load_model
 Organization = load_model('openwisp_users', 'Organization')
 OrganizationOwner = load_model('openwisp_users', 'OrganizationOwner')
 OrganizationUser = load_model('openwisp_users', 'OrganizationUser')
+Group = load_model('openwisp_users', 'Group')
 User = get_user_model()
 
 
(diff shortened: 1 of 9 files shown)


openwisp_users/tests/utils.py
```diff
@@ -76,6 +89,8 @@ def _test_multitenant_admin(self, url, visible, hidden, select_widget=False):
         a superuser can see everything.
         """
         self._login(username='operator', password='tester')
+        if administrator:
+            self._login(username='administrator', password='tester')
```
nemesisdesign, line 93: this is weird here, we either log in as operator or administrator, not both, it may not do anything if we log in first with operator and then with administrator, but it's not clean, so please clean it up

nemesisdesign (changes requested):

openwisp_users/tests/utils.py
codesankalp: Resolved.

PR #326, opened 2022-09-22
[change] Added autocomplete fields in OrganizatinOwner admin #325

### LINKED ISSUE #325: [change] Administrator group should not have permission to change organization owner
Non superadmins should not have permission to change the organization owner from the Organization change page, they should only have view permission.

It would be great if we can improve the readonly result of this part to show the name of the user, because right now in readonly mode the name of the org is shown instead.

pandafy:
Closes #325

### DIFF
--- a/openwisp_users/base/models.py
+++ b/openwisp_users/base/models.py
@@ -221,6 +221,19 @@ class BaseOrganizationUser(models.Model):
                 )
             )
 
+    @property
+    def name(self):
+        """
+        Returns the connected user's full name or user's username if
+        full name method is unavailable (e.g. on a custom user class)
+        or if full name is an empty string.
+        """
+        try:
+            assert self.user.get_full_name() != ''
+            return self.user.get_full_name()
+        except (AssertionError, AttributeError):
+            return str(self.user.username)
+
 
 class BaseOrganizationOwner(models.Model):
     """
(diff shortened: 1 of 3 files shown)


nemesisdesign (changes requested):
I see this:

Screenshot from 2022-09-25 18-25-42

Screenshot from 2022-09-25 18-31-10

It doesn't look what we want right?

It must be because first and last name are empty, I think it may be better to show the username if first and last name are empty.

openwisp_users/base/models.py
```diff
@@ -221,6 +221,19 @@ def clean(self):
...
+        """
+        Returns the connected user's full name or user's username if
+        full name method is unavailable (e.g. on a custom user class)
+        or if full name is an empty string.
+        """
+        try:
+            assert self.user.get_full_name() != ''
+            return self.user.get_full_name()
+        except (AssertionError, AttributeError):
+            return str(self.user.username)
```
nemesisdesign, line 235: why not simply:

```python
return self.user.get_full_name() or str(self.user.username)
```

nemesisdesign (changes requested):

nemesisdesign (commented):
PS: the latest solution works, it works also with the suggested simplification.

openwisp_users/base/models.py
pandafy: I adapted this code from django-organizations. I will update it.

nemesisdesign (approved):

PR #328, opened 2022-09-26
[fix] Fixed user's organization cache invalidation #327

### LINKED ISSUE #327: [bug] Cache invalidation for org owner does not seem to work when owner is changed
I found another bug while debugging https://github.com/openwisp/openwisp-users/issues/325.

Here's how I can replicate it:

- log in as a super administrator
- have some staff user (non superuser) be administrator of an org, let's call this orgadmin
- if you now open a new incognito window, then log in as orgadmin and go to the org page, you should see the possibility of changing the owner
- go back to the superadmin user and change the owner of the org from orgadmin to another user
- go back to the window in which you are logged in as orgadmin and try to change the organization owner

**Expected result:**

The org owner inline should be read only.

**Actual result**:

The org owner inline is still editable.

Looks like a caching invalidation issue.

pandafy:
Invalidate user's organizations cache, if
OrganizationUser.user or OrganizationOwner.organization_user field are changed.

Fixes #327

### DIFF
--- a/openwisp_users/apps.py
+++ b/openwisp_users/apps.py
@@ -2,10 +2,11 @@ import logging
 
 from django.apps import AppConfig
 from django.conf import settings
+from django.contrib.auth import get_user_model
 from django.core.cache import cache
 from django.core.exceptions import ValidationError
 from django.db import IntegrityError, transaction
-from django.db.models.signals import post_delete, post_save
+from django.db.models.signals import post_delete, post_save, pre_save
 from django.utils.translation import gettext_lazy as _
 from openwisp_utils import settings as utils_settings
 from openwisp_utils.admin_theme.menu import register_menu_group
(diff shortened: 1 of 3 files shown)


nemesisdesign (commented):
I do not understand very well this change.
Can you please explain what's the reason behind the bug and how the fix addresses it?

openwisp_users/apps.py
```diff
@@ -109,22 +113,50 @@ def connect_receivers(self):
...
-            user = instance.organization_user.user
+            _invalidate_old_related_obj_cache(instance, 'organization_user')
+
+    @classmethod
+    def _invalidate_user_cache(cls, user):
+        User = get_user_model()
+        if not isinstance(user, User):
+            user = user.user
+
         cache_key = 'user_{}_organizations'.format(user.pk)
```
pandafy, line 158: The user's organizations dict is dependent on user's primary key. When we change the user field of an OrganizationUser, the cache for both the previous user and the new user have to be invalidated. Consider the following scenario:

An OrganizationUser object is created with `OrgA` and `User1` objects. The organization dict of `User1` is updated in the cache. Now, the same `OrganizationUser` object is changed and `User2` is used instead.  The old code only invalidated the organization dict cache for `User2`. Therefore, `User1` organization dict would reflect that it is sill a member of `OrgA`.

nemesisdesign (approved):
I improved the comments and docstrings a bit in c8ca978fd8fbcd912a6f75d3608e6baca697e4b8.
Thanks :+1:

PR #331, opened 2022-10-27
[feature] Added reusable organization filter #270

### LINKED ISSUE #270: [feature] Create a reusable organization filter using select2 and autocomplete
The list of organization accessible to a superadmin user can be potentially very large, this can slow down page loading in all list pages which have a filter by organization.

For this reason we should look for a solution which loads the options asynchronously.

It looks like this is what we need: https://pypi.org/project/django-admin-autocomplete-filter/.

We already have an organization filter which complies with our multi-tenancy requirements in `openwisp_users.multitenancy.MultitenantOrgFilter`, it would be really great if we manage to just update this filter to ship this new feature, that way we won't have to touch other modules at all because those are already using this class.

pandafy:
Closes #270

**Blockers**

- [ ] https://github.com/openwisp/openwisp-utils/pull/322

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,17 +26,15 @@ jobs:
       fail-fast: false
       matrix:
         python-version:
-          - 3.7
-          - 3.8
-          - 3.9
+          - "3.8"
+          - "3.9"
+          - "3.10"
         django-version:
-          - django~=3.0.0
-          - django~=3.1.0
-          - django~=3.2.8
+          - django~=3.2.0
+          - django~=4.0.0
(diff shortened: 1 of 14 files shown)


nemesisdesign (changes requested):
Great progress Gagan!
Let's not forget to change the organization filter in the user list page which even after these changes is still using the old non-autocomplete filter:

Screenshot from 2022-10-28 15-03-19

Any idea on why this coverage test is going down? https://coveralls.io/jobs/108459322/source_files/27171801594

nemesisdesign (approved):

PR #334, opened 2022-12-05
[change] Added `org_id` to the validation of organization endpoint #329

### LINKED ISSUE #329: [change] API /users/organization endpoint POST doesn't return organization ID when org already exists
Current Behavior:

When using POST to set a new ORG in OW and the ORG already exists in OW we get back: 

`{
  "slug": [
    "organization with this slug already exists."
  ]
}`

Expected behavior:

`{
  "slug": [
    "organization with this slug already exists."
  ]
  "org_id: [
     "blahblahblahblah"
  ]
}`

Aryamanz29:
Screenshot from 2023-01-04 16-56-59

- Added `org_id` to the validation when an organization already exists with the given slug.
- Added tests.

Closes #329

### DIFF
--- a/openwisp_users/api/serializers.py
+++ b/openwisp_users/api/serializers.py
@@ -39,7 +39,7 @@ class OrganizationSerializer(ValidatedModelSerializer):
         Custom validation error if an organization
         already exists with the given slug.
         """
-        org = Organization.objects.filter(slug=data.get('slug')).first()
+        org = Organization.objects.filter(slug__exact=data.get('slug')).first()
         if org:
             raise serializers.ValidationError(
                 {
@@ -47,7 +47,7 @@ class OrganizationSerializer(ValidatedModelSerializer):
                     'organization': org.pk,
                 }
             )
-        return data
+        return super().validate(data)
 
 
 class CustomPrimaryKeyRelatedField(serializers.PrimaryKeyRelatedField):
--- a/openwisp_users/api/serializers.py
+++ b/openwisp_users/api/serializers.py
@@ -19,12 +19,6 @@ OrganizationOwner = load_model('openwisp_users', 'OrganizationOwner')
 
 
 class OrganizationSerializer(ValidatedModelSerializer):
-    slug = serializers.SlugField(
-        allow_unicode=False,
-        help_text=_('The name in all lowercase, suitable for URL identification'),
-        max_length=200,
-    )
-
     class Meta:
         model = Organization
         fields = (
@@ -38,6 +32,7 @@ class OrganizationSerializer(ValidatedModelSerializer):
             'created',
             'modified',
         )
+        extra_kwargs = {'slug': {'validators': []}}
 
     def validate(self, data):
         """
@@ -48,8 +43,8 @@ class OrganizationSerializer(ValidatedModelSerializer):
         if org:
             raise serializers.ValidationError(
                 {
-                    'slug': [_('organization with this slug already exists.')],
-                    'org_id': [org.id],
+                    'slug': _('organization with this slug already exists.'),
+                    'organization': org.pk,
                 }
             )
         return data
--- a/openwisp_users/api/serializers.py
+++ b/openwisp_users/api/serializers.py
@@ -32,6 +32,8 @@ class OrganizationSerializer(ValidatedModelSerializer):
             'created',
             'modified',
         )
+        # This allows replacing the default "UniqueValidator" for
+        # the slug field with the custom validation defined below.
         extra_kwargs = {'slug': {'validators': []}}
 
     def validate(self, data):
--- a/openwisp_users/api/serializers.py
+++ b/openwisp_users/api/serializers.py
@@ -19,6 +19,12 @@ OrganizationOwner = load_model('openwisp_users', 'OrganizationOwner')
 
 
 class OrganizationSerializer(ValidatedModelSerializer):
+    slug = serializers.SlugField(
+        allow_unicode=False,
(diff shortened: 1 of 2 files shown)


openwisp_users/api/serializers.py
```diff
@@ -33,6 +39,21 @@ class Meta:
...
+        """
+        Custom validation error if an organization
+        already exists with the given slug.
+        """
+        org = Organization.objects.filter(slug=data.get('slug')).first()
+        if org:
+            raise serializers.ValidationError(
+                {
+                    'slug': [_('organization with this slug already exists.')],
+                    'org_id': [org.id],
```
nemesisdesign, line 52: for consistency with the rest of the API, this should be:

```python
'organization' : org.pk,
```

nemesisdesign (changes requested):
My concern would be that someone would be able to trigger the API endpoint to find out the slug and UUID of other organizations, which would be a security risk, but thinking it more thoroughly I concluded that if a user has the permission to create orgs, they must have access to all orgs as it doesn't make sense to have permission to create orgs and then not be able to access them, it's possible but doesn't make sense. Therefore we could move on with this change.

Please see my suggestion below.

openwisp_users/api/serializers.py
```diff
@@ -19,6 +19,12 @@
 
 
 class OrganizationSerializer(ValidatedModelSerializer):
+    slug = serializers.SlugField(
+        allow_unicode=False,
+        help_text=_('The name in all lowercase, suitable for URL identification'),
+        max_length=200,
+    )
```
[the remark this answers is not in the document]
Aryamanz29: Because this field already has `validators = [UniqueValidator(queryset=Organization.objects.all())>`, I redefined it to use a customised validation error message response. If I recall correctly, and after some thought, I discovered it can avoid this by removing validations by [extra_kwargs (DRF doc)](https://www.django-rest-framework.org/api-guide/serializers/#additional-keyword-arguments) in the `meta` class attribute; see below: -

```py
		# we can use this instead of the redefining serializer field above
        extra_kwargs = {
            'slug': {
                'validators': []
            }
        }
```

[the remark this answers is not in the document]
Aryamanz29: Screenshot from 2023-01-04 16-27-14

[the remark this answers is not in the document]
Aryamanz29: Added in [`12a1435` (#334)](https://github.com/openwisp/openwisp-users/pull/334/commits/12a1435b45e771f9f9fd19337a5757729f800a1d)

openwisp_users/api/serializers.py
```diff
@@ -32,6 +32,22 @@ class Meta:
             'created',
             'modified',
         )
+        extra_kwargs = {'slug': {'validators': []}}
```
nemesisdesign, line 35: what is this doing?

openwisp_users/tests/test_api/test_api.py
```diff
@@ -49,7 +49,7 @@ def test_organization_list_nonsuperuser_api(self):
     def test_organization_post_api(self):
         path = reverse('users:organization_list')
         data = {'name': 'test-org', 'slug': 'test-org'}
-        with self.assertNumQueries(6):
+        with self.assertNumQueries(4):
```
nemesisdesign, line 52: I also do not understand why the queries go down, do you have an idea of why this is happening?

openwisp_users/api/serializers.py
Aryamanz29: @nemesisdesign  I replied here : https://github.com/openwisp/openwisp-users/pull/334#discussion_r1061374455

openwisp_users/tests/test_api/test_api.py
Aryamanz29: After looking deep into the code, I found that in the serializer `validate()` method we are not calling `super().validate()`; therefore, we got 2 queries less than before.

#### Before (with slug `UniqueValidator`):
- `UniqueValidator()` and `super().validate()` both are executing similar queries (total 6 queries).

```sql
2. SELECT (1) AS "a" FROM "openwisp_users_organization" WHERE "openwisp_users_organization"."slug" = 'test-org' LIMIT 1
3. SELECT (1) AS "a" FROM "openwisp_users_organization" WHERE "openwisp_users_organization"."slug" = 'test-org' LIMIT 1
```
Screenshot from 2023-01-12 16-05-39

#### After (with custom slug validation)

- First we validate `slug_field` and then `super().validate()` (total 6 queries).

Screenshot from 2023-01-12 16-16-13

openwisp_users/api/serializers.py
nemesisdesign: ok, can you please summarize this concept in a comment above this line?

Aryamanz29: @nemesisdesign  Added in [`f1eb8ec` (#334)](https://github.com/openwisp/openwisp-users/pull/334/commits/f1eb8ecde45a0c7b1495019fde4c01116c42f315)

openwisp_users/api/serializers.py
```diff
@@ -32,6 +32,8 @@ class Meta:
             'created',
             'modified',
         )
+        # We need to remove the "UniqueValidator" class from the "slug" field
+        # to add our customised validation error message in the API response
```
nemesisdesign, line 36: What about:

This allows replacing the default "UniqueValidator" for
the slug field with the custom validation defined below.

Aryamanz29: done :+1:

nemesisdesign (approved):

PR #335, opened 2022-12-05
[chores] Minor improvements to organization filters

nemesisdesign:
- MultitenantOrgFilter: use all lowercase titles for consistency
- MultitenantRelatedOrgFilter: avoid having to define field_name and title each time

### DIFF
--- a/openwisp_users/multitenancy.py
+++ b/openwisp_users/multitenancy.py
@@ -137,5 +137,4 @@ class MultitenantRelatedOrgFilter(MultitenantOrgFilter):
     one of the organizations the current user is associated with
     """
 
-    parameter_name = None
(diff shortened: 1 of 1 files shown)


openwisp_users/multitenancy.py
```diff
@@ -128,13 +128,14 @@ class MultitenantOrgFilter(AutocompleteFilter):
...
 
 
-class MultitenantRelatedOrgFilter(AutocompleteFilter):
+class MultitenantRelatedOrgFilter(MultitenantOrgFilter):
     """
     Admin filter that shows only objects which have a relation with
     one of the organizations the current user is associated with
     """
 
+    parameter_name = None
```
pandafy, line 140: I don't think this is required.

PR #336, opened 2022-12-12
[skip ci][docs] Updated documentation for using organization admin filters

pandafy:


README.rst
```diff
@@ -1003,14 +1003,43 @@ Admin Multitenancy mixins
...
+        # other attributes
+
 
 * **MultitenantRelatedOrgFilter**: similar ``MultitenantOrgFilter`` but
   shows only objects which have a relation with one of the organizations
-  the current user can manage, this shall be used when the model does not
-  have its own organization field but relies on a parent model which
-  has the organization field.
+  the current user can manage, this shall be used for creating filters
+  for related objects. In the following example, an autocomplete filter is
```
nemesisdesign, line 1025: What about:

```
the current user can manage, this shall be used for creating filters
for related multitenant models.
```

README.rst
```diff
@@ -1003,14 +1003,43 @@ Admin Multitenancy mixins
...
+    class ShelfFilter(MultitenantRelatedOrgFilter):
+        field_name = 'shelf'
+        parameter_name = 'shelf'
+        title = _('Shelf')
+
+    class BookAdmin(admin.ModelAdmin):
+        list_filter = [
+            ShelfFilter,
+        ]
+        # other attributes
```
nemesisdesign, line 1042: Why don't we add examples from other OpenWISP modules and link them? I think it will be a lot easier to understand, we should also provide basic info about the relationships between these models. I believe this example is not easy to understand for a newcomer.

nemesisdesign (changes requested):

nemesisdesign (approved):

PR #338, opened 2022-12-13
[feature] Centralized `ProtectedApiMixin` logic #337

### LINKED ISSUE #337: [chores] Centralize ProtectedApiMixin logic
There's pretty much almost the same logic scattered around in different openwisp modules: https://github.com/search?q=org%3Aopenwisp+ProtectedApiMixin&type=code. 

We should create a reusable class here.

Aryamanz29:
Closes #337

### DIFF
--- a/openwisp_users/api/mixins.py
+++ b/openwisp_users/api/mixins.py
@@ -6,7 +6,7 @@ from rest_framework.exceptions import NotFound
 from rest_framework.permissions import IsAuthenticated
 
 from .authentication import BearerAuthentication
-from .permissions import DjangoModelPermissions
+from .permissions import DjangoModelPermissions, IsOrganizationManager
 
 Organization = swapper.load_model('openwisp_users', 'Organization')
 
@@ -217,8 +217,11 @@ class ProtectedAPIMixin(object):
     Contains authentication and permission classes for API views
     """
 
-    authentication_classes = [BearerAuthentication, SessionAuthentication]
-    permission_classes = [
-        IsAuthenticated,
+    authentication_classes = (
+        BearerAuthentication,
+        SessionAuthentication,
+    )
+    permission_classes = (
+        IsOrganizationManager,
         DjangoModelPermissions,
-    ]
+    )
--- a/README.rst
+++ b/README.rst
@@ -1012,6 +1012,26 @@ Admin Multitenancy mixins
   have its own organization field but relies on a parent model which
   has the organization field.
 
+ProtectedAPIMixin
+~~~~~~~~~~~~~~~~~
+
+This mixin provides a set of authentication and permission classes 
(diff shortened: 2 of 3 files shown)


openwisp_users/api/mixins.py
```diff
@@ -206,3 +210,15 @@ class FilterSerializerByOrgOwned(FilterSerializerByOrganization):
...
+class ProtectedAPIMixin(object):
+    """
+    Contains authentication and permission classes for API views
+    """
+
+    authentication_classes = [BearerAuthentication, SessionAuthentication]
+    permission_classes = [
+        IsAuthenticated,
+        DjangoModelPermissions,
+    ]
```
nemesisdesign, line 224: let's use tuples instead of lists

nemesisdesign (changes requested):
Looks good!

Please add a brief section of this new mixin in the README at the end of the ["Django REST Framework Mixins" section](https://github.com/openwisp/openwisp-users#django-rest-framework-mixins).

See also my comment below.

openwisp_users/api/mixins.py
nemesisdesign: I think this should be:

```python
permission_classes = (
    IsOrganizationManager,
    DjangoModelPermissions,
)
```

It should be this way because looking at the other modules, they all use this except Network Topology.

nemesisdesign (changes requested):

openwisp_users/api/mixins.py
Aryamanz29: > I think this should be:
> 
> ```python
> permission_classes = (
>     IsOrganizationManager,
>     DjangoModelPermissions,
> )
> ```
> 
> It should be this way because looking at the other modules, they all use this except Network Topology.
> 
> Once you update this here, please update the other modules to make sure the `permission_classes` property is not redefined.

sure :+1:

Aryamanz29: > ```python
> permission_classes = (
>     IsOrganizationManager,
>     DjangoModelPermissions,
> )
> ```

I have tested `ProtectedAPIMixin` with these permission classes, and they work in every module except `openwisp-users`, See below :

Screenshot from 2022-12-14 14-12-06

We need to redefined `permission_classes` property only for `openwisp-users` API views.

Aryamanz29: https://github.com/openwisp/openwisp-users/actions/runs/3692932262/jobs/6252339027#step:11:1

nemesisdesign: Thanks for trying this, I started looking into it and I see that we have a lot of repetition and inconsistencies in the different modules. I think we will have to fix it but I need a bit more time to analyze it and come up with a good solution.

nemesisdesign (approved):

Issue #333, opened 2022-12-02
[feature] Make sure we can use autocomplete field for organization relations


nemesisdesign:
We need be able to use autocomplete for the organization fields across OpenWISP.
We have to make sure that when a null value is allowed we continue showing "Shared systewide (no organization)".
Can we add some JS to fix #170 and automatically select the org when there's only 1 value available? "Shared" should be counted too as a valid value, so shared + 1 org is 2 values and it should not be triggered in that case.

PR #339, opened 2022-12-19
[feature] Use autocomplete field for organization relation #333

pandafy:
Closes https://github.com/openwisp/openwisp-users/issues/333

**Blockers**
- [ ] https://github.com/openwisp/openwisp-utils/pull/330


tests/testapp/tests/test_admin.py
```diff
@@ -20,31 +19,6 @@ class TestUsersAdmin(TestOrganizationMixin, TestCase):
         'openwisp_users' if not os.environ.get('SAMPLE_APP', False) else 'sample_users'
     )
 
-    def test_organization_default_label(self):
```
nemesisdesign, line 23: As discussed, let's add some selenium tests to replace these.

nemesisdesign (commented):
Good progress Gagan. I did some early testing.

Something catching my eye is the fact that the length of the dropdown does not equal the length of the field:
Screenshot from 2022-12-19 18-09-18

I was trying an user which only has access to one org:
Screenshot from 2022-12-19 18-11-21
Can we use JS to auto-select the only available org in this case?

openwisp_users/multitenancy.py
```diff
@@ -118,6 +120,13 @@ def multitenant_behaviour_for_user_admin(self, request):
             qs = super().get_queryset(request)
         return qs
 
+    def formfield_for_foreignkey(self, db_field, request, **kwargs):
+        if db_field.name == 'organization':
+            kwargs['widget'] = OrganizationAutocompleteSelect(
+                db_field, self.admin_site, using=kwargs.get('using')
+            )
```
nemesisdesign, line 127: Something I was forgetting, how does this change play out with changes like this one?

https://github.com/openwisp/openwisp-firmware-upgrader/commit/ae64e7f3278e3f5cf07984c072644d2f66fc2bf5#diff-495f69374c163b98300e8e981a9a2bea26f703c3dc960cbfb6a3c08ae1f7b0a1R56

A first look seem not to conflict with it, I guess we have to stop introducing that kind of change (adding organization to the autocomplete fields) in the different modules?

openwisp_users/views.py
```diff
@@ -29,3 +29,13 @@ def get_queryset(self):
         if not self.request.user.is_superuser and org_lookup:
             return qs.filter(**{org_lookup: self.request.user.organizations_managed})
         return qs
+
+    def get_empty_label(self):
+        if self.object_list.model == Organization:
+            return 'Shared systemwide (no organization)'
```
nemesisdesign, line 35: This needs to be flagged as translatable.

nemesisdesign (changes requested):

nemesisdesign (changes requested):
I was testing the interaction of this PR with https://github.com/openwisp/openwisp-radius/pull/463/ and I noticed the following:

Screenshot from 2022-12-19 20-09-47

In the dropdown above, "Shared systemwide" should not be available because this model does not allow it.

If I save without selecting an org I get an exception:

Screenshot from 2022-12-19 20-09-24

openwisp_users/multitenancy.py
pandafy: Yes, I designed this solution in a way that will not require updating other modules of openwisp. Moreover, the default autocomplete fields added by Django won't have "Shared systemwide" option. We should not use it.

pandafy:
@nemesisdesign the `RadiusCheckAdmin` has a bug unrelated to this PR. I was able to replicate that bug on the master branch of openwisp-radius https://github.com/openwisp/openwisp-radius/issues/464

openwisp_users/widgets.py
```diff
@@ -0,0 +1,26 @@
...
+        return reverse('admin:ow-auto-filter')
+
+    def optgroups(self, name, value, attrs=None):
+        groups = super().optgroups(name, value, attrs)
+        if value == [''] and len(groups[0][1]) == 1:
+            groups[0][1].append(
+                self.create_option(
+                    name=name,
+                    value='null',
+                    label=_('Shared systemwide (no organization)'),
```
nemesisdesign, line 20: Let's avoid duplicating the definition of this string which we are using in multiple places, can you please define it in a constant somewhere (eg: multitenancy.py) and change the rest of the code to use the constant? This way we ensure that any change to this value in the future will remain constant across the entire application.

requirements.txt
```diff
@@ -3,6 +3,6 @@ django-extensions~=3.1.0
 django-allauth~=0.51.0
 django-phonenumber-field~=6.0.0
 phonenumbers>=8.11.0,<8.13.0
-openwisp-utils[rest] @ https://github.com/openwisp/openwisp-utils/tarball/master
+openwisp-utils[rest] @ https://github.com/openwisp/openwisp-utils/tarball/autocomplete-view-improvements
```
nemesisdesign, line 6: you can switch these back to master

nemesisdesign (changes requested):

nemesisdesign (approved):

PR #343, opened 2023-02-03
[fix] Fixed bug in UserAdmin when staff user manages multiple orgs

### LINKED ISSUE #324: [bug] User in 2 organizations sees his user listed twice if not superuser
I have the following situation:

* 2 organizations A and B:

grafik

* An user `org_admin` who is admin in both organizations:

grafik

* The superuser `admin` sees the following users:

grafik

* But the user `org_admin` sees the following users:

grafik

The same user is listed twice, it's the same id.

* if the user `org_admin` clicks on one of those two entries, the query fails:

grafik

* if the user `org_admin` has superuser status, everything is fine:

grafik

This happens with openwisp-users 1.0.2.

nemesisdesign:
Without this patch it is possible that the system will return multiple identical results for a staff user who manages multiple orgs.

Fixes #324.

### DIFF
--- a/openwisp_users/multitenancy.py
+++ b/openwisp_users/multitenancy.py
@@ -112,7 +112,7 @@ class MultitenantAdminMixin(object):
                         f'{User._meta.app_label}_organizationuser__organization__in'
                     ): user.organizations_managed
                 }
-            )
+            ).distinct()
             # hide superusers from organization operators
             # so they can't edit nor delete them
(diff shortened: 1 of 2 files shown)


pandafy:
@okraits can you help us test this patch?

nemesisdesign (commented):
I think this solution works. Thanks @pandafy :+1:

PR #352, opened 2023-05-24
[fix] Fixed related org field filtering in multitentant dj-filters

### LINKED ISSUE #351: [bug] django-filters raises FieldError when organization field is specified
When the organization field is not directly related to a model and its relation is specified in the filter lookup, applying the organization filter raises a `FieldError` (as shown below). This may occur because the correct `field_name` is not used in the `FilterDjangoOrganization` class's **filter_for_field** method.

**Steps to reproduce**

1. Update the filter fields of the `LibraryListFilter` class:

```py
class LibraryListFilter(FilterDjangoByOrgManaged):
    class Meta:
        model = Library
        fields = (
            'book',
            # add organization field
            'book__organization',
        )
```

2. Navigate to http://localhost:8000/testing/library/ and apply organization filter.

Screenshot from 2023-05-23 21-18-33

Aryamanz29:
Fixes #351

### DIFF
--- a/openwisp_users/api/mixins.py
+++ b/openwisp_users/api/mixins.py
@@ -225,7 +225,7 @@ class QuerySetRequestMixin(BaseQuerySetRequestMixin):
         # of organizations they're related to
         organization_filter = getattr(user, self._user_attr)
         # if field_name organization then just organization_filter
-        if self.field_name == 'organization':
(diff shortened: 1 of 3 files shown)


nemesisdesign (approved):
Thanks! :+1:

PR #356, opened 2023-07-26
[feature] Added command line utility to export users #355

### LINKED ISSUE #355: [feature] Command line tool to export users data
We need to add a command line tool in OpenWISP Users that exports a CSV with the User model data and any other data related to it.

OpenWISP Users shall provide a list which will store the fields used by this tool, this list will be extended by OpenWISP RADIUS to add more fields, eg:

- id
- username
- email
- password
- first_name
- last_name
- is_staff
- is_active
- date_joined
- phone_number
- birth_date
- location
- notes
- language
- organizations (special case: list openwisp_users_organizationuser relations, for each one show values of organization.name, is_admin, each line separated with \n)
- registered_user.method (defined in openwisp-radius)
- registered_user.is_verified (defined in openwisp-radius)

Fields may be excluded by the user, eg:

```
./manage.py export_users --exclude-fields location,language,registered_user.method
```

We need to watch out from generating too many queries unnecessarily.
We need to use iterator() when looping over all users in the DB.

pandafy:
Closes #355

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -273,6 +273,41 @@ is a valid number of not.
 This allows users to log in by using only the national phone number,
 without having to specify the international prefix.
 
+``OPENWISP_USERS_EXPORT_USERS_COMMAND_CONFIG``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
++--------------+--------------------------+
+| **type**:    | ``dict``                 |
++--------------+--------------------------+
+| **default**: | .. code-block:: python   |
+|              |                          |
+|              |    {                     |
+|              |    'fields': [           |
+|              |        'id',             |
+|              |        'username',       |
+|              |        'email',          |
+|              |        'password',       |
+|              |        'first_name',     |
+|              |        'last_name',      |
+|              |        'is_staff',       |
+|              |        'is_active',      |
+|              |        'date_joined',    |
+|              |        'phone_number',   |
+|              |        'birth_date',     |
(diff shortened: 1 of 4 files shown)


nemesifier (changes requested):
Can you fix the test coverage here?

README.rst
```diff
@@ -1114,10 +1149,32 @@ Admin Multitenancy mixins
...
+Example usage:
+
+.. code-block:: shell
+
+    ./manage.py export_users
+
+**Arguments**:
+
+- `--exclude-fields`: Comma-separated list of fields to exclude from the export.
+- `--filename`: Filename for the exported CSV, defaults to "openwisp_exported_users.csv".
```
nemesifier, line 1169: Please provide an example with the arguments, eg:

```
./manage.py export_users
 --filename export-excluded-pwd-id.csv --exclude-fields password,id
```

README.rst
```diff
@@ -1114,10 +1149,32 @@ Admin Multitenancy mixins
...
+
+You can change the exported fields using the `OPENWISP_USERS_EXPORT_USERS_COMMAND_CONFIG
+<#openwisp_users_export_users_command_config>`_ setting.
```
nemesifier, line 1172: I am not sure I understand the point of this since users can use the `--exclude-fields` flag.
I think it's worth to point out that this setting is meant to be used for advanced customizations.

nemesifier (changes requested):

nemesifier (approved):

PR #361, opened 2023-10-21
fix: Organization.add_user INTEGRITY error

### LINKED ISSUE #360: [bug] Organization.add_user INTEGRITY error
Try using the add_user method of the Organization model to add an user to an organization which does not have a user yet, eg:

```python
org = Organization.objects.create(...)
user = User.objects.create(...)
org.add_user(user)
```

Expected result: works without errors.
Actual result: 

```
django.db.utils.IntegrityError: UNIQUE constraint failed: openwisp_users_organizationowner.organization_user_id
```

We have a receiver in OpenWISP Users which automatically creates an organization owner if a user is added to an organization which doesn't have any owner. Apparently the `add_user` method tries to do that again which triggers this exception.

mohebmithani:
since this project uses 'add_user' method of 'django-organizations' package, I catched the specific integrity error so that I don't interfere with the imported package internals.

Fixes #360

### DIFF
--- a/openwisp_users/base/models.py
+++ b/openwisp_users/base/models.py
@@ -206,6 +206,22 @@ class BaseOrganization(models.Model):
     class Meta:
         abstract = True
 
+    def add_user(self, user, is_admin=False, **kwargs):
+        """
+        We override this method from the upstream dependency to
+        skip the creation of the organization owner, which we perform
+        automatically via a signal receiver.
+        Without this change, the add_user method would throw IntegrityError.
+        """
+
+        if not self.users.all().exists():
+            is_admin = True
+
+        OrganizationUser = load_model('openwisp_users', 'OrganizationUser')
+        return OrganizationUser.objects.create(
+            user=user, organization=self, is_admin=is_admin
+        )
+
 
 class BaseOrganizationUser(models.Model):
     """
--- a/openwisp_users/tests/test_models.py
+++ b/openwisp_users/tests/test_models.py
@@ -334,9 +334,9 @@ class TestUsers(TestOrganizationMixin, TestCase):
         with self.subTest('Test user first and last names are empty'):
             self.assertEqual(str(org_user), f'{user.username} ({org.name})')
 
-    def test_add_user_method(self):
+    def test_add_user(self):
         org = self._get_org()
         user = self._create_user()
         org_user = org.add_user(user)
-
         self.assertIsInstance(org_user, OrganizationUser)
+        self.assertTrue(org_user.is_admin)
--- a/openwisp_users/tests/test_models.py
+++ b/openwisp_users/tests/test_models.py
@@ -333,3 +333,10 @@ class TestUsers(TestOrganizationMixin, TestCase):
 
         with self.subTest('Test user first and last names are empty'):
             self.assertEqual(str(org_user), f'{user.username} ({org.name})')
+
+    def test_add_user_method(self):
+        org = self._get_org()
+        user = self._create_user()
+        org_user = org.add_user(user)
+
+        self.assertIsInstance(org_user, OrganizationUser)


mohebmithani:
@nemesifier This project uses the functionality of `django-organizations` package and if you look into the internals of that package you'll find that `add_user` uses a protected method. So I thought it is better to handle exception then create a user_add method again. Let me know if this is the right approach. If this is right then I can further create a test case for this fix if you want

openwisp_users/models.py
```diff
@@ -25,6 +26,29 @@ class Organization(BaseOrganization, AbstractOrganization):
...
+        overriding this method of ``AbstractOrganization``
+        to handle double organization owner creation
+        """
+        try:
+            org_user = super().add_user(user, **kwargs)
+        except IntegrityError as e:
+            # if ``super().add_user()`` again tries to create
+            # organization owner then catch the error and returns 
+            # the organization user created
+            if "UNIQUE constraint failed: openwisp_users_organizationowner.organization_user_id" in str(e):
```
nemesifier, line 41: I am not sure this would work in all databases

nemesifier (changes requested):
Thanks @mohebmithani! 

A failing test is missing though. When we fix bug we write a test case which replicates the issue first (eg: fails without the fix). Without that we cannot be confident that 1. the bug exists, 2. has been solved, 3. it won't come back in the future due to inadverted changes to the code.

mohebmithani:
@nemesifier yes i justed wanted to verify the approach. I'll implement the test case.

openwisp_users/models.py
mohebmithani: Do you want me to just except integrity error not check the exact message?

nemesifier: I just thought maybe we could find a better way, see here:

https://github.com/bennylope/django-organizations/blob/2dd044956cf03dad9c8c4954ba325d2769fe3643/src/organizations/abstract.py#L96C1-L117

What if we override the `create` method of our org owner model to verify whether an owner already exists before creating a new one?

nemesifier (changes requested):

openwisp_users/models.py
mohebmithani: Overriding was my first solution but I thought may be we wanted to use django-organizations methods that's why i implemented this. Let me override this and send you PR, thanks.

mohebmithani:
@nemesifier I thought about overriding `create` method of org owner. I don't think it would be a good solution. Because `create` method would then check if owner exists or not, If exists then it won't create another object. 

Let's say if someone wanted to add some feature which depends on the object creation of `owner`? if `objects.create` is successful then do **this** otherwise do **that**. I don't think create method should be changed just for this usecase. This error is occuring in `add_user` method then I think only this method should be changed. 

I've changed the implementation of `add_user` a bit. Now it will just create `org user` leaving creation of `org owner` to the signals. I've implemented test as well and tested every test cases. Let me know if more changes are required.

mohebmithani:
@nemesifier just topping up this message as a reminder. Actually i've selected this issue as part of 'Hacktoberfest' and I would realy appreciate if you can look into this before the event ends. Apologies and Thanks in advance

openwisp_users/models.py
```diff
@@ -25,6 +26,24 @@ class Organization(BaseOrganization, AbstractOrganization):
     class Meta(AbstractOrganization.Meta):
         swapper.swappable_setting('openwisp_users', 'Organization')
 
+    def add_user(self, user, **kwargs):
```
nemesifier, line 29: please move this to `openwisp_users/base/models.py`.

openwisp_users/models.py
```diff
@@ -25,6 +26,24 @@ class Organization(BaseOrganization, AbstractOrganization):
...
+        """
+        overriding this method of ``AbstractOrganization``
+        to remove organization owner creation logic as it is
+        being handled separately by a signal of this project.
```
nemesifier, line 33: Let's shorten this docstring, something like the following will work:

We override this method from the upstream dependency to
skip the creation of the organization owner, which we perform
automatically via a signal receiver.
Without this change, the add_user method would throw IntegrityError.

openwisp_users/models.py
```diff
@@ -25,6 +26,24 @@ class Organization(BaseOrganization, AbstractOrganization):
...
+        """
+
+        users_count = self.users.all().count()
```
nemesifier, line 36: since we are at it, we can use `exists()` which is more efficient, will return `True` or `False`.

openwisp_users/models.py
```diff
@@ -25,6 +26,24 @@ class Organization(BaseOrganization, AbstractOrganization):
...
+        if users_count == 0:
+            is_admin = True
+        
+        OrganizationUser = swapper.load_model('openwisp_users', 'OrganizationUser')
+        org_user = OrganizationUser.objects.create(
+            user=user, organization=self, is_admin=is_admin
+        )
```
nemesifier, line 43: let's just return here, eg:

```suggestion
        return OrganizationUser.objects.create(
            user=user, organization=self, is_admin=is_admin
        )
```

openwisp_users/models.py
```diff
@@ -25,6 +26,24 @@ class Organization(BaseOrganization, AbstractOrganization):
...
+
+        return org_user
+
```
nemesifier, line 46:
```suggestion
```

openwisp_users/tests/test_models.py
```diff
@@ -333,3 +333,15 @@ def test_organization_user_string_representation(self):
 
         with self.subTest('Test user first and last names are empty'):
             self.assertEqual(str(org_user), f'{user.username} ({org.name})')
+
+    def test_add_user_method(self):
+        """
+        test adding user does not throw `INTEGRITY` error due to double owner creation
+        logic in `add_user` method and a signal trigger.
```
nemesifier, line 340: Please remove this docstring as it's a redundant with the docstring of the `add_user` method, we do not need it (we can track the reason for this change by looking at the commit history and looking at this PR).

openwisp_users/models.py
```diff
@@ -1,5 +1,6 @@
 import swapper
 from django.contrib.auth.models import Group as AbstractGroup
+from django.db import IntegrityError
```
nemesifier, line 3: Please make sure to run `openwisp-qa-format` and then `./run-qa-checks` to make sure the QA checks are passing.

nemesifier (changes requested):
Thanks @mohebmithani, see my comments below.

mohebmithani:
@nemesifier Thanks for the update. I've incorporated all the feedbacks and have ran all the checks.

openwisp_users/base/models.py
```diff
@@ -206,6 +206,22 @@ def __str__(self):
...
 
+    def add_user(self, user, is_admin=False, **kwargs):
+        """
+        We override this method from the upstream dependency to
+        skip the creation of the organization owner, which we perform
+        automatically via a signal receiver.
+        Without this change, the add_user method would throw IntegrityError.
+        """
+
+        if self.users.all().exists():
```
nemesifier, line 217: Are you sure of this?
Wasn't the core flagging the user as admin if there were no user for the org?
While here it's flagging the user as admin if there's already users for the org.

Shouldn't it be `if not self.users.all().exists()`?

nemesifier (changes requested):

openwisp_users/base/models.py
mohebmithani: @nemesifier really sorry for this silly mistake. Have add not in the if statement.

mohebmithani:
@nemesifier did you get a chance to look at it? hopefully now everything will be on thier place.

nemesifier (approved):

nemesifier:
Thanks @mohebmithani :+1:

PR #363, opened 2023-10-27
[feature] Added password expiration #359

### LINKED ISSUE #359: [feature] Password expiration
We need to implement an optional feature which allows to implement password expiration for end-users.

## Define settings to enable or disable this feature

- Setting and admin UI field to enable/disable feature (and specify how many months), disabled by default
- Setting and admin UI field to enable/disable feature (and specify how many months) also for staff users
- Add fields to user model: password updated date field, allowed to be None, shall remain None for users who use social login or SAML
- Add filter in admin to show password expired users according to the setting
(the filter shall be shown only if the setting is enabled, it's going to be a boolean and if true it must look for users which have the password update date older than the months specified in the settings)

## Changes in allauth views

Other changes needed to allow expired users to change their password:

- `LOGIN_REDIRECT_URL` must point by default to a new view with these characteristics:
  - if the password has not expired and the user is `is_staff=True`, redirect to admin
  - if the password has expired or if it has not expired and the user is not staff (`is_staff=False`), redirect to change password

## Celery tasks

We have discussed adding a background task for celery beat running daily at 1:00 AM by default.

Look for users (use use `iterator()` ORM method) who are member of at least one of the enabled organizations and:

- have a usable password (hence excluding social login / SAML)
- have the password updated date older than X
- have a verified email address - (since we need to send emails)
- are not staff user (is_staff=False) if the feature is disabled for staff users

If the password is about to expire in exactly 7 days, send an email notification with a warning that the password is about to expire and a call to action link to change the password.

These changes will require [updating ansible-openwisp2 to include this task to the celery beat settings](https://github.com/openwisp/ansible-openwisp2/issues/449).

## Other changes

This is how we planned the implementation of this feature:

- Override `User.set_password` method to update the `password_updated` date field if feature is enabled

- Modify auth backend so that (if the feature is enabled) it checks whether the user has a usable password and if the password is expired returns false (adds also a warning message with the message framework which says "Your password has expired, please update your password" with a link to the password change view of allauth).
Add another alternative auth backend, eg: `UsersAllowExpiredPassBackend`, which we'll use in the allauth accounts adapter for the non-admin pages, this backend shall authenticate users also if their password is expired.

- Add backend attribute to allauth adapter which points to the `UsersAllowExpiredPassBackend`

# Other modules

Ansible OpenWISP 2: https://github.com/openwisp/ansible-openwisp2/issues/449
OpenWISP RADIUS: https://github.com/openwisp/openwisp-radius/issues/491
WiFi Login Pages: https://github.com/openwisp/openwisp-wifi-login-pages/issues/713

pandafy:
Closes #359


openwisp_users/admin.py
```diff
@@ -194,7 +194,7 @@ class UserAdmin(MultitenantAdminMixin, BaseUserAdmin, BaseAdmin):
     add_form = UserCreationForm
     form = UserChangeForm
     ordering = ['-date_joined']
-    readonly_fields = ['last_login', 'date_joined']
+    readonly_fields = ['last_login', 'date_joined', 'password_updated']
```
nemesifier, line 197: This is always shown but the field seems to be populated only when the feature is enabled.

openwisp_users/base/models.py
```diff
@@ -84,6 +89,13 @@ def _get_pk(obj):
             raise ValueError('expected UUID, str or Organization instance')
         return str(pk)
 
+    def set_password(self, *args, **kwargs):
+        if app_settings.USER_PASSWORD_EXPIRATION or (
+            self.is_staff and app_settings.STAFF_USER_PASSWORD_EXPIRATION
+        ):
+            self.password_updated = timezone.now().date()
```
nemesifier, line 96: what if we just populate it anyway, that do we have to lose?

openwisp_users/migrations/0019_user_password_updated.py
```diff
@@ -0,0 +1,20 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("openwisp_users", "0018_allow_operator_view_organisation"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="user",
+            name="password_updated",
```
nemesifier, line 15: do we need a data migration to populate this the first time?

openwisp_users/admin.py
nemesifier: For some reason I can't see the `password_updated` field, can you?
Question: how difficult would it be to show this field below the main password field?

nemesifier (changes requested):
Great progress @pandafy :clap:

I tried the admin flow and tried updating the password by inputting the same password and I was able to do this.
I think I did not specify this in the original issue, maybe we forgot to take into account this important detail, but we shouldn't allow to change the password by setting the same password, it's not going to be compliant with the requirement unfortunately.

See my comments below, I will need to do more rounds of review and testing.

openwisp_users/admin.py
pandafy: Screenshot from 2023-10-30 18-15-34

openwisp_users/migrations/0020_populate_password_updated_field.py
```diff
@@ -0,0 +1,21 @@
+# Generated by Django 4.2.2 on 2023-10-18 10:13
+
+from django.contrib.auth import get_user_model
+from django.db import migrations
+from django.utils import timezone
+
+
+def populate_password_updated_field(apps, schema_editor):
+    User = get_user_model()
+    User.objects.update(password_updated=timezone.now().date())
```
nemesifier, line 10: If users do not have a password probably we should avoid this.

tests/openwisp2/settings.py
```diff
@@ -71,6 +71,12 @@
     'django.middleware.clickjacking.XFrameOptionsMiddleware',
 ]
 
+AUTH_PASSWORD_VALIDATORS = [
+    {
+        "NAME": "openwisp_users.password_validation.PasswordReuseValidator",
+    },
+]
```
nemesifier, line 76: Please include the default validators.

nemesifier (changes requested):
It shouldn't be hard to cover these lines by instantiating the adapter class directly and calling the method we need to call to cover the lines: https://coveralls.io/builds/63631301/source?filename=openwisp_users%2Faccounts%2Fadapter.py#L28

Same for this line:
https://coveralls.io/builds/63631301/source?filename=openwisp_users%2Fbackends.py#L70

tests/openwisp2/settings.py
pandafy: This would only create hurdles during development, i.e. we would need to use complex password during development.

nemesifier (changes requested):
When logging in as a non staff user and having the password expired, can't we redirect straight to `/accounts/password/change/`?

nemesifier (changes requested):

nemesifier (changes requested):
Please check my latest changes and rebase on the latest master and review my changes.

nemesifier (approved):

PR #364, opened 2023-11-10
[change] Enable OrganizationUserAdmin by default

nemesifier:
If OrganizationUserAdmin is disabled, `OrganizationOwnerInline` was loading a select with all the users in the DB, which doesn't work on large databases.

This change enables this admin page by default and avoids loading the `OrganizationOwnerInline` if `OrganizationUserAdmin` is disabled.


pandafy (commented):
LGTM!

PR #373, opened 2024-03-11
[feature:api] Allow creating users with a verified email via REST API

nemesifier:
Allow creating users with a verified email via REST API.


pandafy (approved):
LGTM!

PR #374, opened 2024-04-02
[fix] Fix has_password_expired when password_update is None

nemesifier:
The `password_updated` field can be `None`, so trying to perform an addition on it before ensuring the value is not `None` can lead to uncaught exceptions, this patch fixes this issue.


nemesifier:
The tests are failing because of [[bug] Test test_password_expiration_mail is suddenly failing](https://github.com/openwisp/openwisp-users/issues/375).

PR #384, opened 2024-06-06
[fix:tests] Fixed flaky expected date in test_password_expiration_mail

nemesifier:
Build the expected date in the same way the django template framework does.


pandafy (approved):
Thank you for fixing this @nemesifier

PR #389, opened 2024-07-05
[tests] Improved TestUsersAdmin.test_delete_selected_overridden_action_perms

pandafy:
Avoid generating related objects on post_save signal for User object.


nemesifier (approved):
That's a lot better, thanks

PR #444, opened 2025-05-07
[feature] Allowed read only access of shared objects to non-superusers #238

pandafy:
## Checklist

- [x] I have read the [OpenWISP Contributing Guidelines](http://openwisp.io/docs/developer/contributing.html).
- [x] I have manually tested the changes proposed in this pull request.
- [ ] I have written new test cases for new code and/or updated existing tests for changes to existing code.
- [ ] I have updated the documentation.

## Reference to Existing Issue

Closes #238
Closes #448 

**TODOS**
-[x]  DRF integration 
-[x] Tests

PR #477, opened 2025-12-03
[api/mixins] Fix filtering of unrelated fields in FilterSerializerByOrganization #319

### LINKED ISSUE #319: [bug] `FilterSerializerByOrganization` filters fields without organization
`FilterSerializerByOrganization` filters queryset of all related fields using the `organization_lookup`. But, it might be possible that a model  does not have relation to `Organization` (not even an indirect one, e.g. [WiFiClient in openwisp-monitoring](https://github.com/openwisp/openwisp-monitoring/pull/382#discussion_r898221497)). 

We should allow to exclude fields from the organization filtering.

Stoiicc:
This pull request fixes the issue described in #319.
The FilterSerializerByOrganization mixin was applying organization-based filtering to every serializer field that exposed a queryset, regardless of whether the related model actually contained an organization relation.This behavior was inconsistent with the expected logic of the mixin and caused unintended filtering in modules where certain related models are not tied to organizations.

This patch ensures that queryset filtering is applied only to fields whose related model actually contains an organization relation.Fields that do not have such a relation are left untouched.The behavior for fields pointing directly to the Organization model remains unchanged, and the existing include_shared logic and filtering rules for non-superusers continue to work as before.

Fixes #319.

### DIFF
--- a/openwisp_users/api/mixins.py
+++ b/openwisp_users/api/mixins.py
@@ -148,7 +148,9 @@ class FilterByParentOwned(FilterByParent):
 
 class FilterSerializerByOrganization(OrgLookup):
     """
-    Filter the options in browsable API for serializers
+    Filter serializer related-field querysets based on the organizations the
+    current user is allowed to access.
+
     """
 
     include_shared = False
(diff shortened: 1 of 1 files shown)


nemesifier (changes requested):
Thanks @Stoiicc for tackling #319. The underlying goal is valid: the mixin should not apply organization filtering to related fields whose model has no organization relation. But this is a security-critical multitenancy mixin, and the rewrite introduces a pattern I cannot accept as-is.

**Blocker: fail-open exception handling**

The new code wraps the queryset filtering in broad `except Exception: pass`:

```python
try:
    field.queryset = queryset.filter(conditions).distinct()
except Exception:
    pass
```

If anything goes wrong building or applying the filter, this silently leaves the field's queryset unfiltered, i.e. exposing every organization's objects in the browsable API and writable choices. For a mixin whose entire job is org isolation, that is a fail-open data leak. There are several such blocks. Please make filtering fail closed: if the intended filter cannot be applied, the field should end up empty or raise, never wide open. Also replace the broad excepts with the specific exceptions you actually expect.

**Other concerns:**

1. **Auto-detecting org fields is a heuristic.** `_get_org_related_fields` treats any FK to `Organization` as the filter field and ORs across all of them. A model with more than one Organization relation (or an indirect one) could be filtered on the wrong field. Please confirm this matches the real models in play and add tests for the multi-relation case.
2. **Overlap with #444.** PR #444 heavily rewrites this same `api/mixins.py`. The two will conflict and overlap conceptually. Let's sequence them: #444 is the canonical multitenancy/shared-objects work, so I would rebase this on top of (or fold it into) #444 rather than land them independently.
3. This needs comprehensive tests proving org isolation is preserved for every case (the #319 scenario, the org field itself, shared objects, and crucially that no path leaves a field unfiltered).

Right problem, but the fail-open handling is a blocker and it must be coordinated with #444.

PR #490, opened 2026-03-03
[fix] OpenWisp DependencyLoader should be after django template loaders

asmodehn:
## Checklist

- [X] I have read the [OpenWISP Contributing Guidelines](http://openwisp.io/docs/developer/contributing.html).
- [X] I have manually tested the changes proposed in this pull request.
- [ ] I have written new test cases for new code and/or updated existing tests for changes to existing code.

Closes #489

## Description of Changes

Correct order of template loaders in tests settings : OpenWisp DependencyLoader must come after django template loaders to ensure correct templates are found when extending an app.


openwisp_users/templates/admin/login.html
```diff
@@ -1,51 +1 @@
...
-    <a href="{% url 'account_reset_password' %}">Forgot Password?</a>
-  </div>
-  <div class="submit-row">
-    <input type="submit" value="Log in" />
-  </div>
-</form>
-
-</div>
-{% endblock %}
+{% extends 'admin/openwisp_users/login.html' %}
```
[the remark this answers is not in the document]
asmodehn: Hmm I didn't experience this in my local setup... but what do you mean exactly by "make it extend Django's admin template directly" ?

asmodehn:
@nemesifier, @pandafy  Please have a look at this, and related issues, when you can.
In my opinion, this is the proper way to manage template extensions in openwisp.

This prevents finding the wrong base template when django's app loader look for it from an extension, and this way, we can keep the loader order logical, and consistent on all modules, once similar changes are made in other repos.

PS: I ignored all coderabbit comments, since I just moved the file without changes.

tests/openwisp2/settings.py
```diff
@@ -89,14 +89,11 @@
...
         "BACKEND": "django.template.backends.django.DjangoTemplates",
-        "DIRS": [
-            os.path.join(os.path.dirname(BASE_DIR), "openwisp_users", "templates")
-        ],
         "OPTIONS": {
             "loaders": [
                 "django.template.loaders.filesystem.Loader",
-                "openwisp_utils.loaders.DependencyLoader",
                 "django.template.loaders.app_directories.Loader",
+                "openwisp_utils.loaders.DependencyLoader",
```
nemesifier, line 97: I verified this is the important behavioral change. With `SAMPLE_APP=1`, Django app templates need to win before `DependencyLoader`, otherwise `admin/login.html` and `admin/action_confirmation.html` are picked from `openwisp_users` instead of `sample_users`. I would add a tiny regression test for the resolved template origins, because content checks can still pass when the wrong app owns the path.

openwisp_users/templates/admin/login.html
nemesifier, line 1: This proxy template is the right direction in my opinion. It keeps the public override point as `admin/login.html`, but moves the reusable OpenWISP implementation to a namespaced template so it does not fight Django admin's own template resolution.

tests/openwisp2/sample_users/templates/admin/login.html
```diff
@@ -0,0 +1 @@
+{% extends 'admin/openwisp_users/login.html' %}
```
nemesifier, line 1: This is the key part for custom apps. The sample app now owns `admin/login.html` and delegates explicitly to the OpenWISP base template, which is much clearer than relying on loader order tricks. :+1:

tests/openwisp2/sample_users/templates/admin/action_confirmation.html
```diff
@@ -0,0 +1 @@
+{% extends 'admin/openwisp_users/action_confirmation.html' %}
```
nemesifier, line 1: Same here, this makes the intended extension path explicit and gives us a clean place to test that the sample app wins template resolution. :clap:

nemesifier (commented):
I did a quick but deep pass because I don't have time to manually chase every edge.

I verified the issue is real: with `SAMPLE_APP=1` on current master, `admin/login.html` and `admin/action_confirmation.html` resolve to `openwisp_users`, not `sample_users`. On this branch they resolve to `sample_users` first and then delegate to `admin/openwisp_users/...`, so the direction looks right to me.

I would not merge as-is yet: docs still tell users to put `DependencyLoader` before `app_directories.Loader`, and I would add a tiny regression test for template origins.

PS: I merged this branch with the latest master, feel free to rebase if you prefer.

tests/openwisp2/settings.py
asmodehn: done

tests/openwisp2/sample_users/templates/admin/action_confirmation.html
asmodehn: i did that in the tests override in sample_users in the following commit

openwisp_users/templates/admin/login.html
asmodehn: Ok lets go with this then. I ll have a look at other places like this when I get some time (I think controller/pki probably has a similar issue)

nemesifier (approved):
Thanks @asmodehn 🙏

nemesifier:
@coderabbitai resolve

PR #492, opened 2026-03-12
[fix] Use UUID converters for users API detail routes #487

czarflix:
## Checklist

- [x] I have read the [OpenWISP Contributing Guidelines](https://openwisp.io/docs/dev/developer/contributing.html).

Closes #487.

## Description of Changes

This fixes a router-to-model PK mismatch in the users API.

The user and organization detail routes were defined with `str` converters even though the underlying resources are UUID-backed, which allowed malformed values like `not-a-uuid` to resolve to DRF views and return `401` instead of failing at the routing boundary with `404`.

This change updates the five affected routes to use `uuid` converters and adds a focused regression test to assert malformed UUID paths return `404`.

Because the route converter change makes `kwargs["pk"]` a UUID object, this also updates the self-password permission check in `ChangePasswordView` to compare UUID values directly. Without that follow-up, self password changes would incorrectly take the stricter permission path and fail existing API tests.

Manual validation performed:
- `python tests/manage.py test openwisp_users.tests.test_api.test_views --parallel 2`
- `python tests/manage.py test openwisp_users.tests.test_api --parallel 2`
- `python tests/manage.py test openwisp_users.tests.test_api.test_api.TestUsersApi.test_organization_detail_api openwisp_users.tests.test_api.test_api.TestUsersApi.test_get_user_detail_api openwisp_users.tests.test_api.test_api.TestUsersApi.test_change_password_of_superuser_by_superuser openwisp_users.tests.test_api.test_api.TestUsersApi.test_get_email_list_api openwisp_users.tests.test_api.test_api.TestUsersApi.test_put_email_update_api --parallel 2`
- `./run-qa-checks`

No manual documentation update was needed because this change corrects the route definition itself, which is the source of truth for the generated API schema/docs.

## Screenshot

N/A (backend/API change)


czarflix:
@nemesifier all checks passing since upstream fix

nemesifier (commented):
Thanks @czarflix, this is a correct and clean fix for #487. Switching the detail routes from `<str:pk>` to `<uuid:pk>` means a non-UUID id returns a clean 404 instead of a 500, and it lets you simplify the self-password check since both are now UUID objects. The regression test for invalid-UUID -> 404 is the right addition.

Minor:

- Double-check there are no other places in the user/org API views that compared `self.kwargs["pk"]` as a string (the password check was one; a grep for `kwargs["pk"]` will confirm there are no remaining `str(...)` comparisons that now always evaluate false).
- The `email_id` staying `<int:...>` is correct (allauth EmailAddress pk is an int).

Good to merge once that quick grep is confirmed.

nemesifier (approved):
I finally could double check and it's good to merge, thanks!

PR #517, opened 2026-05-12
[feature] Added support for user account expiration #499

### LINKED ISSUE #499: [feature] Support user expiration with automatic deactivation and notification emails
Currently, there is no way to set an expiration date on user accounts. Automatically making users inactive after a defined period or on a set date would allow for more flexible account management and help align with various compliance or business requirements.

- Add an optional, nullable `expiration_date` (`DateField`) to the User model.
- Ensure model validation that enforces `expiration_date` must be set to today or a future date, not in the past.
- Existing users with `expiration_date=None` should continue to behave as they do now and never expire.
- Expose `expiration_date` in both the Django admin and REST API so it can be set for all users. Only users which have the "change" permission for the User model can update this field. 
- Implement a daily periodic task using celery beat which is executed at `00:01` AM, using Django timezone-aware utilities, to:
  - fetch all active users whose `expiration_date` is in the past
  - set `is_active=False` for these users
  - avoid touching users who are already inactive
  - remain idempotent and safe for multiple runs

> [!IMPORTANT]  
> Double check that the celery beat is executed based on the TZ defined in the Django project settings independent of time configured on the OS. 

- Implement reminder email notifications:
  1. Send a reminder email before expiration (default: 7 days prior; make this configurable via Django settings)
  2. Send an expiration/deactivation email when the user is marked inactive
- Add tests to cover at least:
  - model validation for past and future dates
  - automatic deactivation of expired users
  - idempotent repeated task execution
  - reminder email dispatch timing
  - expiration/deactivation email dispatch
  - behavior for users with `expiration_date=None`

**Additional context**
- Use `DateField` instead of `DateTimeField` to stay consistent with the current implementation in `openwisp-radius`.
- Handle all date checks with Django timezone-aware utilities and compare against the local current date used by the application.
- The scheduled task should only deactivate users that are currently active, so it does not interfere with other user state transitions that may already have set `is_active=False`.
- The feature should work alongside existing signals, tasks, and user lifecycle operations already present in `openwisp-users`.
- Suggested settings names could follow the existing naming convention, for example:
  - `OPENWISP_USERS_EXPIRATION_REMINDER_DAYS`
- Migration behavior should be backward compatible: existing rows should default to `NULL`, meaning no expiration date is set.

Related issues:

- https://github.com/openwisp/openwisp-radius/issues/706
- https://github.com/openwisp/docker-openwisp/issues/587
- https://github.com/openwisp/ansible-openwisp2/issues/608

pandafy:
## Checklist

Closes #499.

## Description of Changes

Added user account expiration with automated deactivation and notification emails.

- Added `expiration_date` DateField to the User model with past-date validation
- Added `deactivate_expired_users()` classmethod + Celery task that bulk-deactivates users whose expiration date has passed and sends localized deactivation emails
- Added `expiration_reminder_email()` classmethod + Celery task that sends reminder emails `USER_EXPIRATION_WARNING_DAYS` (default: 7) before expiration
- Exposed `expiration_date` in the admin "Account expiration" fieldset and in API serializers with validation
- Added email templates (HTML/text) for both expiration and reminder notices
- Added `OPENWISP_USERS_EXPIRATION_WARNING_DAYS` setting (0 disables reminders)
- Configured Celery beat schedules to run both tasks daily


openwisp_users/base/models.py
```diff
@@ -197,6 +213,133 @@ def _invalidate_user_organizations_dict(self):
...
+
+        This performs a bulk update to set is_active=False for all active users
+        with expiration_date <= today. It returns the number of users
+        deactivated.
+
+        Emails are sent only to affected users who have a verified email
+        address, while all expired active users are deactivated.
+        """
+        expiry_date = localdate()
+        qs = cls.objects.filter(is_active=True, expiration_date__lte=expiry_date)
```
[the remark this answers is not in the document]
pandafy: @coderabbitai why do you think this?

nemesifier (changes requested):
See my inline comment below.

openwisp_users/api/serializers.py
```diff
@@ -235,6 +238,7 @@ class Meta:
             "is_active",
             "is_staff",
             "is_superuser",
+            "expiration_date",
```
nemesifier, line 252: `SuperUserListSerializer.create()` still validates the model during `save()` by calling `instance.full_clean()`, but this path does not convert `django.core.exceptions.ValidationError` to DRF `ValidationError`.

With this new `expiration_date` API field, `POST /users/` with a past date can raise a 500 instead of returning a 400 field error. The detail serializer already handles this conversion on update.

Please move this validation into the serializer validation path, preferably by reusing/adapting `ValidatedModelSerializer`, or catch `DjangoValidationError` here and re-raise DRF `ValidationError`. Please also add create API coverage for a past `expiration_date`.

pandafy:
@coderabbitai full review!

docs/user/settings.rst
```diff
@@ -261,3 +261,18 @@ change their passwords.
...
+
+============ ===========
+**type**:    ``integer``
+**default**: ``7``
+============ ===========
+
+Number of days before a user's ``expiration_date`` when OpenWISP sends an
+account expiration reminder email.
+
+If set to ``0``, reminder emails are disabled.
```
nemesifier, line 278: Also here let's add a link to the new page.

In the new page, let's add a link to this setting.

nemesifier (changes requested):
Let's add a short docs page for this feature, add a mention of the new feature in the intro page (link to the dedicated page). Let's explain the use case: it's either for contractors who are in charge of specific parts of the network but they are contracted only for a specific period of time, so the expiration reflects the expiration date of their contrac, this facilitates the job of the main admins who don't have to constantly police the accounts of the contractors. The other use case is for [temporary wifi access created via openwisp-radius](https://openwisp.io/docs/dev/radius/user/generating_users.html), we can add an internal link to that page.

openwisp_users/settings.py
```diff
@@ -64,6 +64,9 @@ def export_organizations(user):
 STAFF_USER_PASSWORD_EXPIRATION = getattr(
     settings, "OPENWISP_USERS_STAFF_USER_PASSWORD_EXPIRATION", 0
 )
+USER_EXPIRATION_WARNING_DAYS = getattr(
+    settings, "OPENWISP_USERS_EXPIRATION_WARNING_DAYS", 7
+)
```
pandafy, line 70: @nemesifier what do you think about this?

nemesifier:
@coderabbitai full review

openwisp_users/base/models.py
```diff
@@ -182,6 +201,20 @@ def clean(self):
             raise ValidationError(
                 {"email": _("User with this Email address already exists.")}
             )
+        if self.expiration_date and self.expiration_date < localdate():
```
nemesifier: I think this validation still leaves a dangerous edge case: it allows an existing past `expiration_date` to remain unchanged, which is needed for editing unrelated fields, but it also allows changing `is_active` from `False` back to `True` while keeping that already-expired date. The bulk admin action clears only dates strictly before `localdate()`, while the task treats `expiration_date <= localdate()` as expired, so a user expiring today can also be reactivated with a date the task will immediately consider expired. I think we should reject `is_active=True` when `expiration_date <= localdate()` unless the same update clears or extends the date, and add a regression test for this.

openwisp_users/base/models.py
```diff
@@ -197,6 +230,176 @@ def _invalidate_user_organizations_dict(self):
...
+        # queryset would no longer match these users.
+        deactivated_user_ids = list(
+            cls.objects.filter(
+                is_active=True, expiration_date__lte=expiry_date
+            ).values_list("id", flat=True)
+        )
+        if not deactivated_user_ids:
+            return 0
+
+        users_to_deactivate = cls.objects.filter(id__in=deactivated_user_ids).only(
```
nemesifier: This re-query only uses the ID snapshot captured earlier. If an admin extends/removes the expiration date, or another worker runs the same task, between the snapshot and this loop, we can still deactivate and notify a user that no longer matches `is_active=True, expiration_date__lte=expiry_date`. Two concurrent workers can also select the same users and send duplicate deactivation emails. I think this should re-check eligibility immediately before saving, or process rows under a transaction/row lock such as `select_for_update(skip_locked=True)` where supported, and send emails only for rows actually changed by this task.

openwisp_users/tests/test_models.py
```diff
@@ -494,6 +547,606 @@ def test_password_expiration_mail_settings_disabled(self):
...
+                ],
+            )
+            self._assert_email_recipient(
+                deactivate_expired_users,
+                None,
+            )
+
+    @patch.object(app_settings, "USER_EXPIRATION_WARNING_DAYS", 7)
+    def test_expiration_reminder_email(self):
+        reminder_date = now().date() + timedelta(days=7)
```
nemesifier: The implementation uses `localdate()` for validation, reminders, and deactivation, but these tests use `now().date()`. In the test project `TIME_ZONE` is `Europe/Rome` with `USE_TZ=True`, so there is a daily window where UTC `now().date()` differs from the local date used by the code under test. I think all account-expiration test fixtures should use `localdate()` (and explicit local dates when freezing time), otherwise we can get flaky tests and miss local/UTC off-by-one regressions.

docs/user/account-expiration.rst
```diff
@@ -0,0 +1,32 @@
...
+
+How it works
+------------
+
+When an ``expiration_date`` is set on a user account, OpenWISP Users can
+send reminder emails before the account expires. The reminder period can
+be configured using :ref:`OPENWISP_USERS_EXPIRATION_WARNING_DAYS
+<openwisp_users_expiration_warning_days>`.
+
+Once the expiration date is reached, the account is automatically
```
nemesifier: One question: is this feature enabled by default in ansible-openwisp2? If so, we don't need to say much here.

nemesifier (commented):
Thanks for the great effort here, this is nearly there. I fixed the locale compilation issue locally by correcting the Russian and Slovenian plural headers and verified both files with `msgfmt -c`; the remaining points are mostly about making the solution more robust around edge cases. Please read the inline comments below.

docs/user/account-expiration.rst
pandafy: It is enabled by default in ansible-openwisp2

pandafy:
@nemesifier are you okay with having the Account Expiration fieldset after the Permission fieldset? 

image


Or, would you like to move it around?

nemesifier:
I found two small but important consistency issues around reactivating expired users.

First, the deactivation logic treats an account expiring today as expired, but the admin bulk action was only clearing dates strictly in the past. This meant an inactive user with expiration date set to today could be marked active again while keeping a date that the deactivation task would immediately consider expired.

Second, model validation blocked reactivation with the same old expired date, but it still allowed reactivation if the date was changed to today. Since today is also considered expired by the deactivation task, this would still leave us with an active user in an expired state.

I pushed a fix in 2414098. The admin action now clears expiration dates up to and including today before reactivation, and model validation now rejects reactivation unless the expiration date is moved to a future date or cleared. I also added regression coverage for the past date, today date, and future date cases so this behavior stays consistent.

nemesifier:
> @nemesifier are you okay with having the Account Expiration fieldset after the Permission fieldset?
> 
> image
> Or, would you like to move it around?

I added the field to the add user form and moved it up:

image

nemesifier (dismissed):
@pandafy I've done a number of improvements, please review.
From my side it's ready.

nemesifier:
@coderabbitai full review

docs/user/account-and-password-expiration.rst
```diff
@@ -0,0 +1,75 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+When an ``expiration_date`` is set on a user account, OpenWISP Users can
+send reminder emails before the account expires. The reminder period can
+be configured using :ref:`OPENWISP_USERS_EXPIRATION_WARNING_DAYS
+<openwisp_users_expiration_warning_days>`.
+
+Once the expiration date is reached, the account is automatically
+deactivated unless the expiration date is extended or removed.
+
```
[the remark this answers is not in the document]
nemesifier: the installer takes care

openwisp_users/tests/test_models.py
```diff
@@ -24,6 +33,47 @@
...
+
+    def _setup_deactivation_user(self, label):
+        email = f"user{label}@example.com"
+        with freeze_time("2026-05-10 10:00:00+02:00"):
+            return self._create_user(
+                username=f"testuser{label}",
+                email=email,
+                is_active=True,
+                expiration_date=localdate(),
+            )
```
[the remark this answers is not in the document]
nemesifier: we cannot go back in time you know

pandafy (commented):
LGTM! Thanks for the changes @nemesifier

openwisp_users/admin.py
```diff
@@ -269,15 +281,38 @@ def make_inactive(self, request, queryset):
...
-        count = queryset.count()
+        # Clear expired expiration dates before reactivating users.
+        today = localdate()
+        queryset = queryset.filter(is_active=False)
+        expired_count = queryset.filter(expiration_date__lte=today).update(
+            is_active=True, expiration_date=None
+        )
+        count = queryset.filter(
+            Q(expiration_date__isnull=True) | Q(expiration_date__gt=today)
+        ).update(is_active=True)
```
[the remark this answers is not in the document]
nemesifier: @pandafy let's proceed with this change as discussed.

nemesifier (approved):

PR #542, opened 2026-07-14
[feature] Made disabled organizations readonly but deletable #522

### LINKED ISSUE #522: [change] Make disabled organizations readonly but deletable

When an organization is disabled (`is_active = False`), writes against objects related to that organization are not consistently blocked.

Some paths already behave correctly through membership filtering, but that is not a reliable or explicit protection. A few write paths still bypass it, and disabled organizations can also appear in organization selection widgets.

Image

Make disabled organizations a first-class write guard.

Objects related to a disabled organization should:

- remain **readable**
- remain **deletable**
- be blocked from **creation**
- be blocked from **update**

In addition, organization selection fields, autocomplete widgets, and other FK pickers should exclude disabled organizations.

This behavior should be enforced consistently across:

- Django admin
- DRF views and permissions
- model validation for membership and ownership records
- queryset helpers used by organization selectors and autocomplete fields

**Goals**

- make the disabled-organization rule explicit instead of relying on indirect membership filtering
- provide reusable admin and DRF mixins that downstream apps can adopt
- provide a shared helper/queryset pattern for active organizations and memberships
- ensure the rule applies uniformly, including for superusers unless a view explicitly opts out
- keep read and delete operations available so disabled-organization data can still be audited and cleaned up
- keep organization lifecycle operations working, including re-enabling a disabled organization
- surface clear, translatable error messages instead of silent failures or empty querysets

**Deliverables**

- reusable Django admin and DRF mixins for guarding writes on disabled-organization objects
- model-level validation for membership and ownership records
- updates to organization FK/autocomplete querysets so disabled organizations are excluded from selection
- migration of existing openwisp-users call sites to the new mixins/helpers
- clear error messages when a write is blocked

**Notes**

- This issue is about write protection, not authentication or organization lifecycle changes.
- Existing logic that intentionally shows disabled organizations for auditing or lifecycle management should remain available where needed.
- The change should be implemented in a way that downstream OpenWISP apps can reuse.

Reference: [OpenWISP Device Operation Analysis](https://docs.google.com/spreadsheets/d/1teJ3SatCszbAAF8qYc0tphV293xeI0I8ny6Y8sMs7Z8/edit?usp=sharing&utm_source=chatgpt.com)

pandafy:
## Checklist

Closes #522

## Description of Changes

Made disabled organizations read-only but deletable: objects belonging to a disabled organization can still be viewed and deleted, but not created or modified, in both the admin interface and the REST API. This also applies to the organization record itself. 

Organization selection widgets exclude disabled organizations, while admin list filters keep them visible for auditing.

## Screenshots 

**Change admin of a disabled organization**
image

**OrganizationUser for a disabled organization is rendered as readonly in UserAdmin**

image

**OrganizationAdmin is rendered as readonly for disabled org**
image

**Disabled organization is not shown as an option in OrganizationUser add page**
image

**Objects related to a disabled organization is rendered as readonly but allows deletion**
image


pandafy:
@coderabbitai full review!

openwisp_users/tests/test_api/test_api.py
```diff
@@ -136,6 +139,20 @@ def test_patch_disabled_organization_reenable_api(self):
         org1.refresh_from_db()
         self.assertTrue(org1.is_active)
 
+    def test_reenable_disabled_organization_with_field_edit_api(self):
+        org1 = self._get_org()
+        org1.is_active = False
+        org1.save()
+        path = reverse("users:organization_detail", args=(org1.pk,))
+        # re-enabling and editing another field in one request is rejected,
```
pandafy: This should be present in the docstring.

openwisp_users/tests/test_api/test_api.py
```diff
@@ -655,6 +672,27 @@ def test_create_user_organization_users_disabled_org_api(self):
...
+            "username": "rollbackuser",
+            "email": "rollbackuser@test.com",
+            "password": "password",
+            "organization_users": {"is_admin": False, "organization": org1.pk},
+        }
+        with mock.patch.object(
+            OrganizationUser,
+            "full_clean",
+            side_effect=DjangoValidationError("membership boom"),
+        ):
```
pandafy: Instead of using side_effect, a more meaningful test would be to use a disabled organization. Will change the behaviour of the test?

Moreover, we need to test this in a TransactionTestCase.

tests/testapp/tests/test_multitenancy.py
```diff
@@ -1,9 +1,22 @@
...
+from ..admin import ShelfAdmin
+from ..models import Book, Library, Shelf
 from .mixins import TestMultitenancyMixin
 
+User = get_user_model()
+
+
+class LibraryParentAdmin(MultitenantAdminMixin, admin.ModelAdmin):
+    # Library has no organization field; it is reached through its Book parent
+    multitenant_parent = "book"
```
pandafy: Are you sure this is the right place for placing the ModelAdmin? Is this only for testing? 

I'd prefer to move it to the admin.py file, so we can also test this manually.

nemesifier:
@coderabbitai full review

openwisp_users/admin.py
```diff
@@ -98,34 +99,100 @@ class OrganizationOwnerInline(admin.StackedInline):
...
+        so disabled organizations are excluded from the dropdown for everyone,
+        superusers included (the stock ``admin:autocomplete`` endpoint does not
+        filter them). Only replaces the widget when the field is actually an
+        autocomplete field, so that disabling ``autocomplete_fields`` keeps
+        rendering a plain select.
+        """
+        if db_field.name == "organization" and db_field.name in (
+            self.get_autocomplete_fields(request)
+        ):
+            kwargs["widget"] = OrganizationAutocompleteSelect(
```
nemesifier, line 201: In manual testing, the user add page freezes the browser on this branch. The new organization autocomplete is now used by this inline, but `org-autocomplete.js` assumes there is a `select#id_organization` on the page. Inline fields use ids like `id_openwisp_users_organizationuser-0-organization`, so the jQuery selection is empty and the parent-walking loop never reaches a form. Please check the autocomplete initialization for inline usage. A minimal fix could be to skip the script when the expected field is missing, or make it target the actual widget instance instead of a hardcoded id.

openwisp_users/admin.py
```diff
@@ -583,6 +650,29 @@ def has_change_permission(self, request, obj=None):
...
         return super().has_change_permission(request, obj)
 
+    def get_readonly_fields(self, request, obj=None):
+        """
+        A disabled organization can only be re-enabled: every other
+        field becomes readonly (owner unassignment is still possible
+        via the inline's delete action, which does not go through here).
+        """
+        fields = super().get_readonly_fields(request, obj)
+        if obj and not obj.is_active:
```
nemesifier, line 679: In manual testing, disabling an organization makes the main organization fields read-only, but downstream inlines such as config management settings, organization variables, and geo settings remain writable. Please look into whether this can be handled centrally for organization admin inlines, or through the multitenancy admin utilities, instead of requiring each downstream inline to implement the same guard. Inline writes for disabled organizations need test coverage because they are another path for modifying disabled-organization data.

openwisp_users/base/models.py
```diff
@@ -491,6 +491,27 @@ class Meta:
         abstract = True
 
     def clean(self):
+        if self.organization_id and not self.organization.is_active:
```
nemesifier: This guard only checks the submitted organization. If an existing membership belongs to a disabled organization, it can be moved to an active organization and `full_clean()` will not block it. The owner guard below has the same issue. Please reject real changes when either the persisted organization or the submitted organization is disabled, while still allowing no-op validation and deletion.

openwisp_users/api/permissions.py
```diff
@@ -95,6 +95,31 @@ def validate_membership(self, user, org):
...
+    )
+
+    def has_object_permission(self, request, view, obj):
+        if getattr(view, "allow_disabled_organization_writes", False):
+            return True
+        if request.method in SAFE_METHODS or request.method == "DELETE":
+            return True
+        try:
+            organization = self.get_object_organization(view, obj)
+        except AttributeError:
```
nemesifier, line 118: This fails open if `organization_field` is misspelled or the traversal is broken. I would not treat that as a non-organizational object. Either deny access or let the configuration error propagate. Views that intentionally are not organization-bound should opt out explicitly instead of relying on this exception path.

openwisp_users/api/serializers.py
```diff
@@ -110,7 +165,7 @@ def update(self, instance, validated_data):
                     org_owner = OrganizationOwner.objects.create(
```
nemesifier: This saves `OrganizationOwner` before validation. Save receivers run immediately and can repopulate the user's organization cache from a row that can still be rejected and rolled back by `_full_clean_or_raise()`. Please build the instance, validate it, then save it. The same applies to the other owner creation path below.

openwisp_users/multitenancy.py
```diff
@@ -58,7 +61,60 @@ def get_queryset(self, request):
...
+    def has_change_permission(self, request, obj=None):
+        """
+        Objects belonging to a disabled organization stay readable and
+        deletable, but cannot be changed, regardless of the user being a
+        superuser. Subclasses can opt out with
+        ``disabled_organization_write_protection = False``.
+        """
+        if self.disabled_organization_write_protection and obj is not None:
+            organization = self._get_object_organization(obj)
+            if organization is not None and not organization.is_active:
```
nemesifier, line 102: This new write guard runs after queryset scoping has already filtered disabled organizations out for non-superusers through `user.organizations_managed`, which only includes active organizations. As a result, an organization manager who disables an organization loses the ability to see or delete its existing objects. If the intended behavior is superuser-only access after disable, the issue and docs should say so. Otherwise, read and delete scoping needs to include disabled organizations while write choices remain active-only.

openwisp_users/api/serializers.py
```diff
@@ -354,20 +417,16 @@ def update(self, instance, validated_data):
...
-                    str(org_user_data["organization"].id)
-                    in instance.organizations_dict.keys()
-                ):
-                    if org_user.is_admin != org_user_data.get("is_admin"):
-                        org_user.is_admin = org_user_data["is_admin"]
-                        org_user.full_clean()
-                        org_user.save()
-                    else:
-                        org_user.delete()
+                if org_user.is_admin != org_user_data.get("is_admin"):
```
nemesifier: This path can delete an existing membership when `is_admin` is submitted unchanged, and it can raise `KeyError` if `is_admin` is omitted but `.get("is_admin")` differs from the current value. Please make the contract explicit: require `is_admin`, or treat omission as leave unchanged. In either case, do not delete the membership implicitly unless that behavior is intentional and covered by a test.

openwisp_users/tests/test_api/test_api.py
```diff
@@ -807,3 +956,25 @@ def test_expiration_date_none_api(self):
...
+class TestUsersApiTransaction(TestOrganizationMixin, TransactionTestCase):
+    def setUp(self):
+        self.client.force_login(self._get_admin())
+
+    def test_create_user_organization_users_disabled_org_api(self):
+        # A membership validation failure (here, the disabled-organization
+        # guard) after the user row is written must roll the user back
+        # instead of leaving a half-created account behind.
+        path = reverse("users:user_list")
+        org1 = self._create_org(name="disabled-org", is_active=False)
```
nemesifier, line 1033: This test does not exercise the rollback path it describes. `OrgUserCustomPrimarykeyRelatedField.get_queryset()` only returns active organizations, so this request fails during serializer validation before `SuperUserListSerializer.create()` writes the user. Use an active organization for `is_valid()`, then deactivate it before `save()` so the membership validation fails after the user row has been written inside the transaction.

openwisp_users/tests/test_models.py
```diff
@@ -410,6 +410,133 @@ def test_add_user(self):
...
+            org.is_active = False
+            org.save()
+            org_user.is_admin = True
+            with self.assertRaisesMessage(
+                ValidationError,
+                "Memberships of a disabled organization cannot be modified.",
+            ):
+                org_user.full_clean()
+            # deleting the row must still work
+            org_user.delete()
```
nemesifier, line 445: Please store the primary key before calling `delete()`. Django clears `pk` on delete, so this assertion queries `pk=None` and does not prove this row was removed. The owner deletion assertion below has the same issue.

nemesifier (changes requested):
Good progress but incomplete, requires more work and more testing with other key modules.

nemesifier:
Follow-up from manual testing in `openwisp-controller`.

I tested admin objects that belong to a disabled organization. Device admin still has issues:

- `organization` field becomes empty
- `organization` field still looks editable
- Device map tab still looks editable
- Credentials inline still looks editable

Other objects belonging to the same disabled organization looked ok in the same test:

- Template
- VPN server
- Subnet

The reference to `openwisp-controller` does not mean I expect `openwisp-users` to accommodate controller-specific behavior directly. It means this reusable change needs to be tested against key modules before merging, especially `openwisp-controller`, `openwisp-monitoring`, `openwisp-notifications`, and `openwisp-firmware-upgrader`.

If a module-specific admin implementation is bypassing this protection, we should identify the pattern and standardize it as much as possible, so downstream modules do not need to reimplement disabled-organization protection differently.

openwisp_users/api/mixins.py
```diff
@@ -55,6 +66,8 @@ def queryset_organization_conditions(self):
 
     def get_queryset(self):
         qs = super().get_queryset()
+        if getattr(self, "select_related_organization", True):
+            qs = qs.select_related(self.org_field)
```
nemesifier, line 83: `select_related(self.org_field)` runs before object permissions. A misspelled `organization_field` raises `FieldError` while evaluating the queryset, so `DisabledOrgReadOnly` never gets to return `False` and the request ends as HTTP 500. The test at `tests/testapp/tests/test_permission_classes.py:630-647` currently asserts this crash. Please validate the relation before applying this optimization, or avoid it when invalid, then assert 403 and an unchanged object while preserving `select_related()` for valid direct and nested paths.

openwisp_users/multitenancy.py
```diff
@@ -58,7 +61,80 @@ def get_queryset(self, request):
...
+        return organization
+
+    def has_change_permission(self, request, obj=None):
+        """
+        Objects belonging to a disabled organization stay readable and
+        deletable, but cannot be changed, regardless of the user being a
+        superuser. Subclasses can opt out with
+        ``disabled_organization_write_protection = False``.
+        """
+        if self.disabled_organization_write_protection and obj is not None:
```
nemesifier, line 100: This protects object change views, but mutating changelist actions call `has_change_permission(request, obj=None)` and then receive the selected queryset. A superuser can therefore run an update action on disabled-organization rows. `openwisp_controller/config/admin.py:722-749` is a concrete example because `change_group` calls `queryset.update()`. Please reject or remove disabled-organization rows before non-delete actions execute, preserve delete actions and the opt-out, and add an action regression test.

openwisp_users/multitenancy.py
```diff
@@ -67,18 +143,36 @@ def _edit_form(self, request, form):
...
+        keep_disabled_org_pk = None
+        if not self.disabled_organization_write_protection and obj is not None:
+            organization = self._get_object_organization(obj)
+            if organization is not None and not organization.is_active:
+                keep_disabled_org_pk = organization.pk
+        if org_field:
+            allowed = Q(is_active=True)
+            if keep_disabled_org_pk is not None:
+                allowed |= Q(pk=keep_disabled_org_pk)
+            org_field.queryset = org_field.queryset.filter(allowed)
```
nemesifier, line 181: This relation filtering is inside the non-superuser branch, so superusers can still select a shelf, template, CA, VPN, credentials object, or another relation belonging to a disabled organization. Issue #522 explicitly requires other FK pickers to exclude disabled organizations for everyone. Please always restrict these querysets to active organizations or shared objects, then apply the managed-organization restriction for non-superusers. The existing superuser expectation in `test_book_shelf_fk_queryset` also needs to change.

openwisp_users/api/serializers.py
```diff
@@ -174,12 +229,16 @@ def update(self, instance, validated_data):
...
 class OrgUserCustomPrimarykeyRelatedField(serializers.PrimaryKeyRelatedField):
+    default_error_messages = {
+        "does_not_exist": DISABLED_ORGANIZATION_ERROR_MESSAGE,
+    }
+
     def get_queryset(self):
         user = self.context["request"].user
         if user.is_superuser:
-            queryset = Organization.objects.all()
+            queryset = Organization.active.all()
```
nemesifier: Filtering every disabled organization here also prevents deletion of an existing disabled-organization membership. Field validation returns 400 before `SuperUserDetailSerializer.update()` can resolve the membership and reach its deletion branch, and there is no other REST deletion path for it. Please allow only the current user's existing disabled membership to resolve on detail updates. New disabled memberships must remain invalid and role changes must remain blocked. Add a regression test that deletes the existing membership.

openwisp_users/admin.py
```diff
@@ -98,34 +99,100 @@ class OrganizationOwnerInline(admin.StackedInline):
...
+            and not instance.organization.is_active
+        ):
+            org_field = form.fields.get("organization")
+            if org_field is not None:
+                org_field.disabled = True
+                org_field.queryset = Organization.objects.filter(
+                    pk=instance.organization_id
+                )
+            if "is_admin" in form.fields:
+                form.fields["is_admin"].disabled = True
```
nemesifier: This disables only the two built-in membership fields. If a swapped `OrganizationUser` adds a field and a downstream inline exposes it, that field remains writable; `BaseOrganizationUser.clean()` also compares only the built-in fields. I would derive the protected fields from model or form metadata and cover an extended membership model.

Confidence: medium. This requires a downstream swapped model exposing an additional membership field, but swapped-model support is a documented project requirement.

openwisp_users/admin.py
```diff
@@ -577,12 +644,40 @@ def get_inline_instances(self, request, obj=None):
...
+        via the inline's delete action, which does not go through here).
+        """
+        fields = super().get_readonly_fields(request, obj)
+        if obj and not obj.is_active:
+            editable_fields = [
+                f.name
+                for f in self.model._meta.local_fields
+                if f.editable and f.name != "is_active"
+            ]
+            fields = list(fields) + [f for f in editable_fields if f not in fields]
```
nemesifier, line 688: `local_fields` omits editable many-to-many fields added by a swapped Organization model, so those fields remain writable while the organization is disabled. Please include editable local many-to-many fields, or derive the read-only set from the bound form, and add an extended-model regression.

Confidence: medium. The default Organization form has no affected editable M2M field; this concerns supported swapped Organization models.

openwisp_users/api/serializers.py
```diff
@@ -66,7 +82,7 @@ def get_queryset(self):
             queryset = OrganizationUser.objects.filter(
                 Q(organization__in=user.organizations_managed)
             )
-        return queryset.select_related()
+        return queryset.filter(organization__is_active=True).select_related()
```
nemesifier: A normal GET, modify, PUT cycle cannot re-enable a disabled organization that has an owner. The detail response includes that owner, but this queryset removes its membership before `OrganizationDetailSerializer.validate()` runs. Even if it reached `validate()`, the current code marks every submitted owner as changed, including the existing owner. Please allow the existing owner value during re-enable validation, compare it with the persisted owner, and extend the PUT regression with an organization that already has an owner.

openwisp_users/static/openwisp-users/js/org-autocomplete.js
```diff
@@ -7,58 +7,61 @@
...
-      }
+    // Hence, we need to update the value of the selected option before
+    // submission of the form.
+    //
+    // Every organization autocomplete widget (top-level or inline) renders
+    // `data-field-name="organization"`, so we bind to each one instead of a
+    // single hardcoded id.
+    $("select[data-field-name='organization']").each(function () {
+      var orgSelect = $(this);
+      orgSelect.closest("form").on("submit", function () {
```
nemesifier: This callback closes over only the selects present at document ready. Django later clones the empty inline row, but the form callback still reads the original empty-template select, not the newly added one. A dynamically added nullable organization widget can submit literal `null`. Please register one form submit handler that finds all current organization selects at submission time and cover a dynamically added nullable inline.

Confidence: medium. The built-in membership inline is non-nullable; this affects downstream nullable organization inlines using this reusable widget.

openwisp_users/tests/test_admin.py
```diff
@@ -1864,6 +1868,317 @@ def test_only_superuser_can_delete_inline_org_owner(self):
...
+                change_path,
+                {
+                    "user": str(user.pk),
+                    "organization": str(org.pk),
+                    "is_admin": "on",
+                },
+                follow=True,
+            )
+            org_user.refresh_from_db()
+            self.assertFalse(org_user.is_admin)
```
nemesifier, line 2076: Please assert `response.status_code == 403` here. The unchanged database value alone can also result from form validation failure, so it does not prove `has_change_permission()` blocked the POST. CodeRabbit's outstanding suggestion on this test is correct.

nemesifier (commented):
AI-assisted review using GPT-5.6 Sol high. I read the PR description, all review and issue comments, issue #522 and its linked OpenWISP follow-ups, and the complete current diff. I found several remaining gaps in the reusable guards.

One integration issue cannot be placed inline because it is in a downstream repository: `openwisp_controller/config/admin.py:603-607` calls `super().has_change_permission(request)` without forwarding `obj`, so `MultitenantAdminMixin` receives `obj=None` and direct Device edits remain writable after the organization is disabled. I do not expect controller-specific code in this PR, but I need that controller fix and regression coverage before treating the reusable guard as validated.

All current GitHub checks pass, including the full Python and Django matrix. The passing tests do not cover the gaps below, and one test currently codifies the HTTP 500 described inline.

The `AGENTS.md:13-55` contributor-policy rewrite is unrelated to issue #522. Please drop those changes from this branch and handle them separately so this already large security and multi-tenancy change stays focused.

docs/user/basic-concepts.rst
```diff
@@ -149,6 +149,56 @@ instance of the platform.
     `django-organizations
     <https://github.com/bennylope/django-organizations>`_ third-party app.
 
+.. _disabling_an_organization:
+
+Disabling an Organization
+-------------------------
+
+Superusers and managers of the organization can disable it, by unchecking
```
nemesifier: Org managers can disable the org? Can they also enable it again?

pandafy:
@coderabbitai full review!

openwisp_users/api/mixins.py
```diff
@@ -158,29 +188,52 @@ def _user_attr(self):
...
+        """
+        Restrict the querysets of writable relational fields so users can
+        only select active organizations they manage, and related objects
+        belonging to those organizations (including shared objects when
+        ``include_shared`` is set).
+        """
         user = self.context["request"].user
-        # superuser can see everything
-        if user.is_superuser or user.is_anonymous:
+        organization_filter = None
```
nemesifier, line 198: Declared nested serializers are initialized before they inherit the root request context. They have no queryset, so the loop skips them and never reruns after binding. A manager can submit an active managed Book with a nested Shelf in a disabled or unmanaged organization. Reapply `filter_fields()` when a nested serializer is bound, and add a negative API regression that verifies neither object is created.

openwisp_users/base/models.py
```diff
@@ -479,21 +484,37 @@ def __str__(self):
...
+        update_fields = kwargs.get("update_fields")
+        super().save(*args, **kwargs)
+        if (
+            not is_new
+            and (update_fields is None or "is_active" in update_fields)
+            and self.is_active != self._initial_is_active
+        ):
+            signal = organization_enabled if self.is_active else organization_disabled
+            transaction.on_commit(
+                lambda: signal.send(sender=self.__class__, instance=self)
```
nemesifier: Both callbacks close over the same mutable instance. A false then true transition in one atomic block sends `organization_disabled` with `instance.is_active == True`. Updating `_initial_is_active` before rollback also suppresses a retried real transition on the reused instance. Capture immutable transition state for each callback, keep the snapshot rollback-safe, and add transaction coverage for both cases.

openwisp_users/admin.py
```diff
@@ -578,12 +633,36 @@ def get_inline_instances(self, request, obj=None):
 
     def has_change_permission(self, request, obj=None):
         """
-        Allow only managers and superuser to change organization
+        Keep disabled organizations accessible so superusers can re-enable them;
+        read-only fields enforce the remaining restrictions.
         """
         if obj and not request.user.is_superuser and not request.user.is_manager(obj):
             return False
+        if obj and not obj.is_active:
```
nemesifier, line 669: This trusts `user.is_manager(obj)` before returning true for every disabled organization. That lookup uses the two-day membership cache, while status invalidation is queued from `pre_save`. A primed stale cache lets a former manager reach this branch and re-enable the organization. Filter active organizations at the database level for non-superuser scopes, require a superuser here, and enqueue invalidation after commit.

openwisp_users/multitenancy.py
```diff
@@ -58,42 +60,113 @@ def get_queryset(self, request):
...
+                if parent is None:
+                    break
+            organization = getattr(parent, "organization", None)
+        return organization
+
+    def has_change_permission(self, request, obj=None):
+        """
+        Block changes to disabled organizations unless the admin opts out.
+        """
+        if self.disabled_organization_write_protection and obj is not None:
```
nemesifier, line 100: Django authorizes changelist actions with `obj=None`, then hands the selected queryset directly to the action. The guard is skipped here, so an action using `queryset.update()` can modify disabled rows. The pushed `test_disabled_organization_mutating_action_is_blocked` regression demonstrates this. Guard selected objects before non-delete actions run, with an explicit action opt-out where needed.

openwisp_users/api/mixins.py
```diff
@@ -22,6 +33,17 @@ def org_field(self):
...
+    def _organization_relation_is_valid(self, model):
+        for part in self.org_field.split("__"):
+            try:
+                field = model._meta.get_field(part)
+            except FieldDoesNotExist:
+                return False
+            if not (field.concrete and (field.many_to_one or field.one_to_one)):
+                return False
+            model = field.related_model
+        return True
```
nemesifier: A valid to-one path ending at `User` also returns true here. `DisabledOrgReadOnly` then checks that user's `is_active` rather than the object's Organization, silently permitting writes to disabled-organization objects. Require the terminal related model to be the swappable Organization model and make the runtime permission deny a resolved non-Organization object.

openwisp_users/base/models.py
```diff
@@ -508,6 +529,39 @@ class Meta:
...
     def clean(self):
+        original = None
+        if not self._state.adding:
+            original = (
+                self.__class__.objects.select_related("organization")
+                .filter(pk=self.pk)
+                .first()
+            )
+
+        changed = (
```
nemesifier: This hard-codes the built-in membership fields. A swapped `OrganizationUser` with an editable field such as `details` can change that field after its organization is disabled because `changed` remains false. Compare editable concrete model fields instead, and add coverage with the existing sample swapped model.

openwisp_users/base/models.py
```diff
@@ -479,21 +484,37 @@ def __str__(self):
...
         """
-
         if not self.users.all().exists():
             is_admin = True
 
         OrganizationUser = load_model("openwisp_users", "OrganizationUser")
-        return OrganizationUser.objects.create(
-            user=user, organization=self, is_admin=is_admin
-        )
+        org_user = OrganizationUser(user=user, organization=self, is_admin=is_admin)
```
nemesifier: `add_user()` advertises `**kwargs` but drops them here, so extension fields on a swapped membership cannot be initialized through the public API. Pass `**kwargs` to `OrganizationUser(...)` and cover an extension field.

openwisp_users/admin.py
```diff
@@ -98,34 +99,88 @@ class OrganizationOwnerInline(admin.StackedInline):
...
+            and instance.organization_id
+            and not instance.organization.is_active
+        ):
+            org_field = form.fields.get("organization")
+            if org_field is not None:
+                org_field.disabled = True
+                org_field.queryset = Organization.objects.filter(
+                    pk=instance.organization_id
+                )
+            if "is_admin" in form.fields:
```
nemesifier: A downstream inline that exposes an extra swapped membership field leaves it writable because only `organization` and `is_admin` are disabled. Keep deletion enabled, but disable every other form field for an existing membership in a disabled organization.

openwisp_users/admin.py
```diff
@@ -578,12 +633,36 @@ def get_inline_instances(self, request, obj=None):
...
+    def get_readonly_fields(self, request, obj=None):
+        """
+        Lock every field except ``is_active`` while disabled; owner removal uses
+        the inline delete action.
+        """
+        fields = super().get_readonly_fields(request, obj)
+        if obj and not obj.is_active:
+            editable_fields = [
+                f.name
+                for f in self.model._meta.local_fields
```
nemesifier, line 682: This omits editable `local_many_to_many` fields on a swapped Organization, leaving them writable while the organization is disabled. Include editable local many-to-many fields in the readonly calculation.

openwisp_users/api/mixins.py
```diff
@@ -158,29 +188,52 @@ def _user_attr(self):
...
+        organization = getattr(self.instance, "organization", None)
+        if (
+            getattr(view, "allow_disabled_organization_writes", False)
+            and organization is not None
+            and not organization.is_active
+        ):
+            queryset |= field.queryset.filter(pk=organization.pk)
+        if organization_filter is not None:
+            field.allow_null = False
+            queryset = queryset.filter(pk__in=organization_filter)
```
nemesifier: The opt-out branch above adds the current disabled organization, but this non-superuser filter removes it again because `organizations_managed`, `organizations_owned`, and `organizations_dict` contain only active organizations. Preserve the current disabled organization through this second filter for opted-out updates, while still rejecting creation in or reassignment to disabled organizations. Add a manager regression for an intentionally allowed view.

tests/testapp/tests/test_selenium.py
```diff
@@ -137,9 +151,37 @@ def test_shelf_add_form_organization_field(self):
...
+        self.login(username=self.admin_username, password=self.admin_password)
+        self.open(path)
+        WebDriverWait(self.web_driver, 5).until(
+            EC.presence_of_element_located(
+                (By.CSS_SELECTOR, "select[id$='-organization'] + span.select2")
+            )
+        )
+        self.logout()
+
+    def test_dynamic_organization_inline_normalizes_shared_value_on_submit(self):
```
nemesifier, line 262: This creates a synthetic select and dispatches a prevented submit event. It does not exercise Django's add-inline control, a cloned Select2 widget, a real form submission, or persisted `organization=None`. Replace or supplement it with a browser regression that adds a real nullable inline, selects the shared option, submits, and verifies the saved object.

openwisp_users/base/models.py
```diff
@@ -470,6 +471,10 @@ class BaseOrganization(models.Model):
     email = models.EmailField(_("email"), blank=True)
     url = models.URLField(_("URL"), blank=True)
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._initial_is_active = self.is_active
```
nemesifier: Reading `self.is_active` in `__init__()` resolves a deferred field. The pushed `test_deferred_organization_queryset_num_queries` regression shows that evaluating three `Organization.objects.only("id")` rows executes four queries. Initialize without resolving deferred fields, or load the persisted value only when a transition must be evaluated.

openwisp_users/base/models.py
```diff
@@ -479,21 +484,37 @@ def __str__(self):
     class Meta:
         abstract = True
 
+    def save(self, *args, **kwargs):
```
nemesifier, line 484: Lower priority simplification: this duplicates the `is_active` transition detection already performed by `handle_org_is_active_change()` in the `pre_save` receiver. Use these lifecycle signals as the one transition mechanism and connect cache invalidation to them after commit, removing the second query and detector.

tests/testapp/tests/test_permission_classes.py
```diff
@@ -363,3 +371,324 @@ def test_org_user_access_shared_object(self):
...
+            template.refresh_from_db()
+            self.assertEqual(template.organization_id, org.pk)
+
+        with self.subTest("shared object unaffected"):
+            shared_template = self._create_template(
+                name="shared-template", organization=None
+            )
+            shared_url = reverse("test_template_detail", args=[shared_template.pk])
+            response = self.client.put(
+                shared_url,
```
nemesifier, line 453: Lower priority simplification: the superuser-only and manager-only CRUD matrices are strict subsets of the both-role matrix below, while the operations-subset test exercises only helper configuration. The later protected-mixin CRUD test also duplicates the earlier focused test. Keep the both-role, session, opt-out, and parent cases, then remove the duplicate matrices.

openwisp_users/tests/test_models.py
```diff
@@ -1195,3 +1357,63 @@ def test_expiration_reminder_email_recipient_selection(self):
                 expiration_reminder_email,
                 None,
             )
+
+
+class TestOrganizationSignalsTransaction(TestOrganizationMixin, TransactionTestCase):
+    def test_organization_disabled_signal(self):
```
nemesifier: Lower priority simplification: the disable, enable, unrelated-update, `update_fields`, and creation signal tests repeat the same transaction setup and signal contexts. Combine them into one `TransactionTestCase` method with independent `subTest` setup, as required by the repository guidance for related transaction cases.

openwisp_users/tests/test_api/test_api.py
```diff
@@ -116,13 +120,89 @@ def test_organization_patch_api(self):
         self.assertEqual(r.status_code, 200)
         self.assertEqual(r.data["name"], "test org change")
 
+    def test_patch_disabled_organization_field_without_reenabling_api(self):
```
nemesifier: Lower priority simplification: this test and the next two each disable an organization, build the same URL, PATCH, refresh, and assert. Combine the field-only update, re-enable-only update, and re-enable-with-field-update cases into one method with independent `subTest` setup. Keep the PUT and owner cases separate.

nemesifier (changes requested):

openwisp_users/multitenancy.py
```diff
@@ -58,42 +60,113 @@ def get_queryset(self, request):
...
+            organization = self._get_object_organization(obj)
+            if organization is not None and not organization.is_active:
+                return False
+        return super().has_change_permission(request, obj)
```
pandafy, line 104: @nemesifier I need your input for tie-breaking.

Currently, `has_change_permission()` returns `False` for objects belonging to a disabled organization. This correctly prevents editing the object with a `403`, but it also prevents deleting inline objects (e.g. an organization owner).

This is due to Django's admin implementation: when the parent has no change permission, Django does not process the change form POST and disables add/change/delete permissions for its inlines as well. So we cannot get both `403` for the parent and inline deletion through the standard Django admin.

Options:

- **A — Keep the current `403`:** simplest, but inline objects cannot be deleted. This restores the previous behavior.
- **B — Allow the parent change form but make it read-only:** keeps inline deletion working, but requests return `200` instead of `403`. We would also need to hide the Save buttons and success message to avoid misleading users.
- **C — Override Django's admin internals:** keep `403` while allowing inline deletion by changing how Django handles the change form and inline formsets. This gives the desired behavior but adds significant complexity and maintenance cost.

openwisp_users/api/mixins.py
pandafy: @coderabbitai can you explain this review comment?

openwisp_users/multitenancy.py
nemesifier: I double checked all the inlines which are affected by this, and there's no object that we need to delete so badly that justifies adding more complexity, so we can go with A, make sure this decision is embedded into  the docstring. I'll update the issue description to reflect this decision.

AGENTS.md
```diff
@@ -74,6 +74,8 @@ If instructions conflict, repository config and CI workflows win first, official
 - Cached lookups must check permission and organization scope on every request. Changed endpoints need cross-organization regression tests.
 - If you change swapped-model behavior, tenant isolation, auth flows, or admin/API permissions, cover both package-level and integration tests.
 - Changes to HTTP REST API endpoints or Django REST Framework serializers must include tests for permissions, input validation, filtering or pagination when supported, and organization or tenant boundaries where applicable.
+- Objects belonging to a disabled organization must be readable and deletable; creation and updates must be blocked across all relevant write paths.
+- New organization-scoped models and operations, including direct or indirect organization relationships, must follow this contract and include explicit tests. Block non-CRUD operations that change state or perform writes on behalf of the object.
```
pandafy: The other option would be to have a dedicated section for Disabled Organization with more explicit instructions 

### Disabled organizations

For objects belonging to a disabled organization:

* **Allow:** read and delete.
* **Block:** create and update.
* Enforce this across all write paths, including API, admin, model, signals,
  and background tasks.
* New models and operations must follow this contract and include explicit
  tests for allowed and blocked operations.
* Block any other operation that changes state or performs a write on behalf
  of the object.

nemesifier:
@coderabbitai verify adherence to all the rules defined in `AGENTS.md`.

PR #544, opened 2026-07-28
[change] Overhauled password expiration feature to avoid interfering with SSO and breaking REST API #511

### LINKED ISSUE #511: [bug] Password expiration blocks SSO and REST recovery

pandafy:
## Checklist

- [x] I have read the [OpenWISP Contributing Guidelines](https://openwisp.io/docs/dev/developer/contributing.html) and [OpenWISP Anti AI Spam Policy](https://openwisp.io/docs/dev/general/code-of-conduct.html#anti-ai-spam-policy).

Fixes #511

## Description of Changes

Added self-service password management REST API endpoints:

- `POST /api/v1/users/password/reset/`
- `POST /api/v1/users/password/reset/confirm/`
- `POST /api/v1/users/user/password/change/`

These endpoints were previously provided by OpenWISP RADIUS. They are fundamental operations for every OpenWISP installation, including those without OpenWISP RADIUS, so they now live in openwisp-users. The implementation uses [dj-rest-auth](https://github.com/iMerica/dj-rest-auth) and applies the authentication endpoint throttling rate to all password self-service requests.

Redesigned password expiration handling to avoid conflicts with passwordless authentication methods such as SAML, OAuth and sesame magic links. Users authenticated without a local password are no longer asked to update an expired local password. `PasswordExpirationMiddleware` now returns a JSON `403` response with recovery URLs to REST API clients instead of an HTML redirect, which breaks API clients.


openwisp_users/templates/custom_password_reset_email.html
```diff
@@ -0,0 +1,11 @@
+{% load i18n %}{% load l10n %}
+<div class="msg">
+<p>
+{% blocktrans %}You're receiving this email because you requested a password reset for your user account at
+{{ site_name }}.{% endblocktrans %}
+</p>
+<p>{% trans "Your username, in case you've forgotten:" %} {{ user.get_username }}</p>
+<p>
+{% trans "Please click on the button below to open a page where you can choose a new password:" %}
+</p>
```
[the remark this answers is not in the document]
pandafy: @Coderabbitai the call to action button is implemented in openwisp-users

openwisp_users/apps.py
```diff
@@ -137,6 +140,34 @@ def connect_receivers(self):
...
+        )
+
+    @classmethod
+    def handle_password_login(cls, request, **kwargs):
+        set_authentication_method(request, PASSWORD)
+
+    @classmethod
+    def handle_allauth_login(cls, request, sociallogin=None, **kwargs):
+        if sociallogin is not None:
+            set_authentication_method(request, EXTERNAL)
```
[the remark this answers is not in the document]
pandafy: @coderabbitai, i don't understand

openwisp_users/apps.py
```diff
@@ -137,6 +140,38 @@ def connect_receivers(self):
...
+        )
+        allauth_user_logged_in.connect(
+            self.handle_allauth_login,
+            dispatch_uid="openwisp_users_handle_allauth_login",
+        )
+
+    @classmethod
+    def handle_password_login(cls, request, **kwargs):
+        backend = getattr(kwargs.get("user"), "backend", "")
+        if backend == "sesame.backends.ModelBackend":
```
pandafy: We need explicit handling for Sesame backend so we can except users logging in with magic links.

nemesifier (commented):
AI-assisted review (GPT-5.6 Sol / high). The findings have not been manually validated.

I left the outstanding findings inline. The CodeRabbit observation about the test-only password-reset serializer is correct and remains important. The Kilo and CodeRabbit requests to reject token issuance for expired passwords are stale because the latest issue #511 discussion explicitly dropped that requirement. The login-by-code suggestion is too broad for this PR because Sesame magic links are the supported passwordless path here and are explicitly covered. The HTML CTA comment was correctly withdrawn, but the CTA target still returns 405. Other automated comments addressed by the current head are not repeated.

openwisp_users/api/views.py
```diff
@@ -61,6 +67,58 @@ def post(self, request, *args, **kwargs):
         return super().post(request, *args, **kwargs)
 
 
+class PasswordResetView(BasePasswordResetView):
```
nemesifier, line 77: High: This endpoint only uses `PasswordResetSerializer` because the test project overrides `REST_AUTH["PASSWORD_RESET_SERIALIZER"]`. A default installation gets dj-rest-auth's serializer, accepts only `email`, and bypasses the documented username and phone lookup. I would set `serializer_class = PasswordResetSerializer` directly on this view and test it without the test-only override.

openwisp_users/api/views.py
```diff
@@ -191,6 +249,17 @@ def update(self, request, *args, **kwargs):
         )
 
 
+class PasswordChangeView(BasePasswordChangeView):
```
nemesifier: High: The old password is not always verified as the docstring claims. dj-rest-auth defaults `OLD_PASSWORD_FIELD_ENABLED` to `False`, removes `old_password`, and allows an authenticated session or Bearer token to replace the password. The tests hide this with a test-only setting. I would use an OpenWISP-owned serializer which always requires and verifies `old_password`.

openwisp_users/auth.py
```diff
@@ -0,0 +1,83 @@
...
+    Sessions without a marker predate this feature, or never had one to
+    begin with (eg: token-based APIs with no session). Fall back to the
+    user's persisted ``last_login_method``, and finally to PASSWORD, so
+    expiration keeps being enforced when the method is unknown.
+    """
+    method = request.session.get(SESSION_KEY) if request is not None else None
+    if method is not None:
+        return method
+    user = user if user is not None else getattr(request, "user", None)
+    return getattr(user, "last_login_method", "") or PASSWORD
```
nemesifier: High: An unmarked session inherits the user's global `last_login_method`. An external login in another browser can therefore make an older password-authenticated session bypass expiration, while a later password login can misclassify an external session. When a request has no session marker, I would default that request to `PASSWORD` and reserve the persisted fallback for callers with no request. Please cover two independent sessions.

openwisp_users/api/views.py
```diff
@@ -61,6 +67,58 @@ def post(self, request, *args, **kwargs):
...
+
+    def get_password_reset_url(self, user, token):
+        """
+        Points the emailed link at the REST confirm endpoint instead of
+        allauth's HTML page. Isolated here so a subclass (e.g.
+        openwisp-radius) can point it at a different frontend without
+        touching token generation or email sending.
+        """
+        uid = user_pk_to_url_str(user)
+        confirm_path = reverse("users:rest_password_reset_confirm")
```
nemesifier: High: The email tells the user to click a reset button, but this target only accepts `POST`. Clicking the generated URL performs `GET` and returns 405, so the default recovery flow is not usable. Please point the CTA to a GET-capable password selection page, or provide one, and make the test follow the emailed URL.

openwisp_users/middleware.py
```diff
@@ -14,36 +36,75 @@ class PasswordExpirationMiddleware:
         "account_reset_password_done",
         "account_reset_password_from_key",
         "account_reset_password_from_key_done",
+        "change_password",
```
nemesifier: High: I do not think this endpoint can be exempted. It lets a superuser or organization manager change another user's password without presenting their own old password, so an expired privileged session can continue account takeover operations. Please remove this exemption and keep only the self-service recovery route exempt.

openwisp_users/api/views.py
```diff
@@ -61,6 +67,58 @@ def post(self, request, *args, **kwargs):
...
 
 
+class PasswordResetView(BasePasswordResetView):
+    """
+    Requests a password reset e-mail for a user identified by username,
+    e-mail address or phone number, mirroring the identifiers accepted at
+    local login.
+    """
+
+    throttle_classes = [AuthRateThrottle]
```
nemesifier, line 85: Medium: `AuthRateThrottle` subclasses `AnonRateThrottle`, which returns no cache key for authenticated requests. Any logged-in user can bypass this limit and send unlimited reset emails. The same ineffective throttle is used by confirm and password change. Please use throttles that cover authenticated requests and add authenticated-request coverage.

openwisp_users/middleware.py
```diff
@@ -14,36 +36,75 @@ class PasswordExpirationMiddleware:
...
     admin_login_path = reverse_lazy("admin:login")
     admin_index_path = reverse_lazy("admin:index")
-    account_change_password_path = reverse_lazy("account_change_password")
 
     def __init__(self, get_response):
         self.get_response = get_response
 
     def __call__(self, request):
+        session_authenticated_before = AUTH_SESSION_KEY in request.session
+        if session_authenticated_before and self._is_expired_password_session(request):
```
nemesifier: Medium: This check runs before DRF authenticates an accompanying Bearer token. A request with an expired session cookie and a valid existing key is rejected even though protected API views prioritize `BearerAuthentication`. The current test has no session and misses this case. Please cover mixed credentials and ensure only a successfully authenticated Bearer request bypasses the session restriction.

openwisp_users/middleware.py
```diff
@@ -14,36 +36,75 @@ class PasswordExpirationMiddleware:
         "account_reset_password_done",
         "account_reset_password_from_key",
         "account_reset_password_from_key_done",
+        "change_password",
+        API_PASSWORD_CHANGE_URL_NAME,
+        "users:rest_password_reset",
+        "users:rest_password_reset_confirm",
     ]
```
nemesifier, line 44: Medium: The final issue discussion intentionally allows token issuance with an expired password, but an expired session cookie still blocks `users:user_auth_token` here before submitted credentials reach the view. I would exempt that route and add a regression test where the client already has an expired session.

docs/user/rest-api.rst
```diff
@@ -68,26 +68,74 @@ Example usage:
 
     {"token": "7a2e1d3d008253c123c61d56741003db5a194256"}
 
-.. _authenticating_rest_api:
+.. _rest_password_reset:
```
nemesifier: Medium: This replaced the existing `authenticating_rest_api` anchor and Bearer-token usage section instead of adding the recovery documentation after it. `docs/developer/django-rest-framework-utils.rst` still references the removed anchor. Please restore that section and keep the new material alongside it.

docs/user/rest-api.rst
```diff
@@ -105,6 +153,28 @@ Change User password
...
+.. code-block:: text
+
+    POST /api/v1/users/password/reset/confirm/
+
+Change Own Password
+~~~~~~~~~~~~~~~~~~~
+
+.. code-block:: text
+
+    PUT /api/v1/users/user/password/change/
```
nemesifier: Low: This summary still advertises `PUT` and `PATCH`, but the view only implements `POST`. Please make it match the corrected detailed section.

openwisp_users/api/views.py
```diff
@@ -61,6 +69,59 @@ def post(self, request, *args, **kwargs):
         return super().post(request, *args, **kwargs)
```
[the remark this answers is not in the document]
nemesifier: This no longer applies. The latest issue #511 discussion explicitly dropped rejection at token issuance because downstream clients need to obtain a token to complete password recovery. Allowing issuance is intentional. The remaining bug is that an expired session cookie causes the middleware to reject this endpoint before it validates the submitted credentials.

AI-assisted (GPT-5.6 Sol / high), not manually validated.

openwisp_users/middleware.py
pandafy: @nemesifier, I think the current implementation is sane. If we can block token issuance with an expired password we should block it. 

In openwisp-radius, it is more of a problem because the downstream application (WiFi Login Pages) will use the Bearer Token to change the password and it wouldn't be able to do so without obtaining token. 

However, the approach is contradicting and make the maintenance harder. 

A user that  has once gained the Token can perform all operations with the REST API even with an expired password.

openwisp_users/api/views.py
pandafy: @nemesifier this is how it's done in openwisp-radius right now. 

https://github.com/openwisp/openwisp-radius/blob/6c4baf41e0216bca15635fac554c71fe8db83979/tests/openwisp2/settings.py#L308

And, this is something which will be configured by either [ansible-openwisp2](https://github.com/openwisp/ansible-openwisp2) and [docker-openwisp](https://github.com/openwisp/docker-openwisp/issues/660)

openwisp_users/middleware.py
pandafy: @nemesifier I don't think this is a real concern.

pandafy:
@coderabbitai full review!

tests/openwisp2/sample_users/migrations/0001_initial.py
```diff
@@ -33,6 +33,23 @@ class Migration(migrations.Migration):
...
+                            " obtained using the local password. When false, the"
+                            " token came from an external method (eg: SSO, SAML)"
+                            " and password expiration is not enforced for it. None"
+                            " means no token has been issued for this user since"
+                            " this feature was introduced."
+                        ),
+                        null=True,
+                        verbose_name="password based token",
+                    ),
+                ),
```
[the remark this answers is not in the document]
nemesifier: This is an sample app for automated testing / example purposes to demonstrate how OpenWISP users can be extended and customized. In order to keep this clean, we strive to not create additional migrations. This app is not meant to be maintained.

openwisp_users/tests/test_middlewares.py
```diff
@@ -43,7 +48,162 @@ def test_queries_middleware_present(self):
...
+        # the final #511 decision allows token issuance even with an
+        # expired password, so a browser holding an expired-password
+        # session must still be able to reach the token endpoint.
+        self._login_expired_admin()
+        response = self.client.post(
+            reverse("users:user_auth_token"),
+            data={"username": "admin", "password": "tester"},
+        )
+        self.assertEqual(response.status_code, 200)
+        self.assertIn("token", response.json())
```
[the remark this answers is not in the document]
nemesifier: We concluded that there's no sane way around this for the time being.

nemesifier:
@coderabbitai review

openwisp_users/base/models.py
```diff
@@ -76,6 +76,22 @@ class AbstractUser(BaseUser):
         default=settings.LANGUAGE_CODE,
     )
     password_updated = models.DateField(_("password updated"), blank=True, null=True)
+    # A user can only have one Token at a time. Storing this flag on the user avoids
+    # introducing a companion model for Token, keeping the implementation simpler and
+    # eliminating additional queries to determine whether the token is password-based.
+    password_based_token = models.BooleanField(
```
nemesifier, line 83: It looked like this field was not being used in this module, but after checking how openwisp-radius uses this, I understood why it's needed. Some authentication flows issue a token after SAML, social or passwordless authentication. Later requests only contain that token, so the Django session marker is no longer available. Persisting how the token was obtained is necessary in those cases.

What makes me uncomfortable is the current public API:

```python
record_password_based_token(user, True)
record_password_based_token(user, False)
```

This asks downstream applications to understand the boolean and update internal OpenWISP Users state correctly. Every token-issuing view must remember which value to write and when to write it. It is also easy to create a token and forget to update the corresponding state, or to update the state without replacing the token.

I think we can provide a safer and easier API from OpenWISP Users. For example:

```python
token = issue_auth_token(request, user, renew=False)
```

The exact name is not important, but the helper should own the complete operation: determine how the request was authenticated, create or renew the DRF token, store whether it was password-based, and return the token.

OpenWISP Users already has the information needed for the current flows:

- A request that has just validated local credentials is password-based.
- An authenticated Django session can use the existing session marker.
- Sesame authentication is implemented here and can be recognized as passwordless.
- Social login is already recorded as passwordless in the session.

A companion read API such as `is_password_based_token(token)` would let downstream applications consume this information without accessing the field directly.

This keeps the rules in one place and gives downstream applications a straightforward public API. It also preserves the current functionality without asking each integration to understand how OpenWISP Users stores the result.

Please also update the field and helper comments to explain that this information is persisted for downstream stateless authentication flows where the original Django session is no longer available.

nemesifier (changes requested):
I found a and fixed a number of seemingly minor issues. Please consult the commit history to double check my changes. I had to rebase on master due to some changes in AGENTS.md that affected this PR.

There's 2 remaining changes to handle:

1. This PR is not a pure bug fix. We have mixed a number of changes: new REST API endpoints (a feature), plus changes to a new feature which I am not sure how to handle in the release notes. The problem is that if we squash merge this, we'll likely fail to communicate the new REST API endpoints for password management in this module.
Resolving this will require to rewrite the history in this PR. We can do this as last step once the other remaining issue is resolved, please focus on the next comment.

nemesifier:
@coderabbitai one final review please.

openwisp_users/api/urls.py
```diff
@@ -49,18 +49,18 @@ def get_api_urls(api_views=None):
...
-                views.rest_password_reset_confirm,
-                name="rest_password_reset_confirm",
+                views.password_reset_confirm,
+                name="user_password_reset_confirm",
             ),
             path(
                 "users/user/password/change/",
                 views.password_change,
-                name="rest_password_change",
+                name="user_password_change",
```
pandafy: @nemesifier the URL names does not reflect that they are REST related URLs. 

Just something to point out that we have other views that deals with password change. 

```
./tests/manage.py show_urls | grep password/change
/accounts/password/change/	openwisp_users.accounts.views.PasswordChangeView	account_change_password
/accounts/password/change/success/	django.views.generic.base.TemplateView	account_change_password_success
/api/v1/users/user/password/change/	openwisp_users.api.views.PasswordChangeView	users:user_password_change
```

openwisp_users/api/views.py
```diff
@@ -265,12 +268,17 @@ def update(self, request, *args, **kwargs):
...
     """
     Self-service password change endpoint for authenticated users.
     """
 
     throttle_classes = [AuthRateThrottle]
+    # Intentional difference from ``ProtectedAPIMixin``: permission_classes
+    # here need to check only that the user is authenticated and nothing
+    # else, but authentication_classes are reused.
+    permission_classes = (IsAuthenticated,)
+    serializer_class = PasswordChangeSerializer
```
pandafy, line 270: FYI, the PasswordChangeView in openwisp-radius only allows bearer authentication.

openwisp_users/middleware.py
```diff
@@ -64,6 +69,20 @@ def __call__(self, request):
...
+        except Resolver404:
+            return False
+        view_class = getattr(resolver_match.func, "cls", None)
+        if view_class is None or not issubclass(view_class, APIView):
+            return False
+        authorization = request.headers.get("Authorization", "")
+        return (
+            authorization.lower().startswith("bearer ")
+            and BearerAuthentication in view_class.authentication_classes
+        )
```
pandafy, line 84: This does not validate whether the token is valid or not. Token validation is deferred to the view.

pandafy:
@coderabbitai full review!

openwisp_users/tests/test_api/test_views.py
```diff
@@ -34,7 +71,38 @@ def test_invalid_uuid_routes_return_404(self):
             "/api/v1/users/user/not-a-uuid/email/1/",
         )
 
-        for path in invalid_uuid_paths:
-            with self.subTest(path=path):
-                response = self.client.get(path)
+        for url_path in invalid_uuid_paths:
+            with self.subTest(path=url_path):
+                response = self.client.get(url_path)
                 self.assertEqual(response.status_code, 404)
```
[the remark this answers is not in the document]
nemesifier: Not needed here. I updated the rule on master.

openwisp_users/api/views.py
nemesifier: Overall consistency is more important than replicating things 1:1. When in doubt, apply this rule.

openwisp_users/middleware.py
nemesifier: I don't understand if you're bringing up a problem and suggesting to change something or just making a statement. Please clarify.

openwisp_users/api/urls.py
nemesifier: Yes, you're right, I had noticed something along this line. The existing view names are inconsistent with what we use in other modules where we generally prefix REST API view names with `api_`. I think the best thing to do is to prefix these new views with `api_` in this PR and deal with the inconsistent naming in future PRs.

nemesifier: I changed my mind, for now let's keep this naming, which is consistent in this module, the change addressing the global consistency between different modules is going to come after this PR is merged. There's a good number of inconsistencies that have to be handled and this is not the right PR to handle it.

openwisp_users/auth.py
```diff
@@ -0,0 +1,135 @@
...
+    if user is None or not user.is_authenticated:
+        return None
+    if renew:
+        Token.objects.filter(user=user).delete()
+    token, _created = Token.objects.get_or_create(user=user)
+    password_based = _is_password_based_request(request, user)
+    user.password_based_token = password_based
+    get_user_model().objects.filter(pk=user.pk).update(
+        password_based_token=password_based
+    )
```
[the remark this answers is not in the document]
nemesifier: Thanks for raising the provenance ambiguity. After reviewing the actual callers and credential lifecycle, we decided not to rotate the DRF token implicitly when provenance changes. The generic API key is intended to remain stable unless the caller explicitly passes `renew=True`; rotating it on authentication-method transitions would unexpectedly invalidate API clients and could cause repeated token churn when users alternate between local and external authentication. Existing social and SAML captive-portal flows already pass `renew=True` where renewal is intentional. The underlying issue is the coupling between the generic DRF token, user-level provenance, and the RADIUS/WiFi Login Pages handoff. We plan to address that separately with a more appropriate credential-exchange design rather than changing the generic API key lifecycle in this PR. The implicit-rotation change and its tests have therefore been removed, while explicit renewal behavior remains unchanged.

openwisp_users/middleware.py
pandafy: I was only making a statement. We don't check if the token was created using expired password.

nemesifier:
@coderabbitai any outstanding issue?

openwisp_users/base/forms.py
```diff
@@ -0,0 +1,107 @@
...
+                context["call_to_action_url"] = url_generator(request, user, token)
+            context.update(extra_email_context)
+            self.send_mail(
+                subject_template_name,
+                email_template_name,
+                context,
+                from_email,
+                user.email,
+                html_email_template_name,
+            )
```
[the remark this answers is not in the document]
nemesifier: We are not going to suppress mail backend exceptions here. In production, the configured e-mail backend generally schedules delivery through Celery, so scheduling failures must remain visible. Returning the normal success response when scheduling or sending failed would falsely tell the user that a reset e-mail was sent and would hide an operational error. We prefer an explicit failure over silent success; a user-friendly error response can be considered separately if needed.

openwisp_users/base/forms.py
```diff
@@ -0,0 +1,107 @@
...
+        if token_generator is None:
+            # Must match the token generator dj-rest-auth's
+            # PasswordResetConfirmSerializer uses to validate the token
+            # (allauth.account.forms.default_token_generator when allauth
+            # is installed), otherwise every confirm request is rejected.
+            if "allauth" in settings.INSTALLED_APPS:
+                from allauth.account.forms import default_token_generator
+            else:
+                from django.contrib.auth.tokens import default_token_generator
+            token_generator = default_token_generator
```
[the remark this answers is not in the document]
nemesifier: `django-allauth[socialaccount]` is a required dependency of this package, and the no-social configuration removes only `allauth.socialaccount`. Rather than retain an unsupported optional-allauth fallback, the latest commit imports and always uses allauth’s token generator, which is the same generator used by confirmation.

nemesifier:
@coderabbitai review

nemesifier:
@coderabbitai:

1. allauth is a hard dependency, the case in which it's not installed was removed, thanks for pointing this out
2. replied in the original thread with the full response
3. and 4. are handled

nemesifier:
@coderabbitai resume

nemesifier:
@coderabbitai resolve

nemesifier (approved):
It's ready.

PR #555, opened 2026-08-08
[feature] Added REST API endpoints for organization memberships #543

### LINKED ISSUE #543: [feature] Add dedicated REST API endpoints for managing user organization memberships

Currently, there is no dedicated REST API endpoint to manage organization memberships (add, remove, or modify). Users must be added/modified through the superuser user detail endpoint (`PUT /api/v1/users/user/{id}/`), and removal requires changing the `is_admin` value in a way that is undocumented and unintuitive.

Add dedicated REST API endpoints to manage organization membership:

- `GET /api/v1/users/user/{user_id}/organization-membership/`
- `POST /api/v1/users/user/{user_id}/organization-membership/`
- `PUT/PATCH /api/v1/users/user/{user_id}/organization-membership/{org_id}/`
- `DELETE /api/v1/users/user/{user_id}/organization-membership/{org_id}/`

Both approaches should support proper permission checks (organization managers, superusers only).

**Describe alternatives you've considered**

1. Continue using the indirect approach via the user detail endpoint with undocumented behavior
2. Document the existing behavior (changing `is_admin` to same value deletes membership) and consider it sufficient

**Additional context**

**Current Issues:**
- No clear way to remove a user from an organization via REST API
- The workaround (sending same `is_admin` value to delete) is undocumented and counter-intuitive
- Only superusers can manage memberships through the current endpoint
- Missing organization manager permissions checks for membership management
- The `OrganizationUserSerializer` exists in the codebase but is only used internally in the superuser serializer

**Benefits:**
- Clearer, more intuitive API for organization membership management
- Better permission control (allow org managers to manage memberships)
- Improved documentation and discoverability
- Consistency with REST API best practices
- Enables programmatic management of organization memberships

BHARATH0153:
## changes

- Added dedicated REST API endpoints to manage user organization memberships

- Memberships can be managed by superusers and by organization managers of the organizations involved; managers are restricted to the organizations they manage and cannot manage superusers.

Closes #543

## Checklist

- [x] I have read the [OpenWISP Contributing Guidelines](https://openwisp.io/docs/dev/developer/contributing.html) and [OpenWISP Anti AI Spam Policy](https://openwisp.io/docs/dev/general/code-of-conduct.html#anti-ai-spam-policy).


BHARATH0153:
@nemesifier  @pandafy once check this whenever you are free thanks!

pandafy:
@coderabbitai full review!

openwisp_users/api/views.py
```diff
@@ -331,6 +332,72 @@ def get_object(self):
...
+        filter_kwargs = {
+            "organization_id": self.kwargs["org_id"],
+        }
+        obj = get_object_or_404(queryset, **filter_kwargs)
+        self.check_object_permissions(self.request, obj)
+        return obj
+
+    def update(self, request, *args, **kwargs):
+        kwargs["partial"] = True
+        return super().update(request, *args, **kwargs)
```
nemesifier, line 406: [High] Deleting an organization owner membership raises `organizations.exceptions.OwnershipRequired` from `OrganizationUser.delete()`. DRF does not handle this exception, so a valid `DELETE` returns 500. Please catch it in the destroy path, return a translated 400 validation response, and add a regression test for deleting the owner membership.

openwisp_users/api/views.py
```diff
@@ -331,6 +332,72 @@ def get_object(self):
...
+        return context
+
+
+class OrganizationMembershipListCreateView(
+    BaseOrganizationMembershipView, ListCreateAPIView
+):
+    pagination_class = OpenWispPagination
+
+    def get_queryset(self):
+        return super().get_queryset().filter(user_id=self.kwargs["pk"])
```
nemesifier, line 385: [Low] During schema introspection, `swagger_fake_view` has no `pk` in `kwargs`. The base method returns an empty queryset, but this line still evaluates `self.kwargs["pk"]` and raises `KeyError`, which drf-yasg catches after logging a traceback. Please short-circuit here as `EmailListCreateView` does and add schema regression coverage.

nemesifier (changes requested):
I found two issues that need to be addressed before this can be merged. The tenant scoping itself looks effective.

openwisp_users/api/views.py
```diff
@@ -252,6 +253,72 @@ def get_object(self):
...
+        return self.get_organization_queryset(qs)
+
+    def get_organization_queryset(self, qs):
+        orgs = self.request.user.organizations_managed
+        app_label = User._meta.app_config.label
+        filter_kwargs = {
+            "is_superuser": False,
+            f"{app_label}_organizationuser__organization_id__in": orgs,
+        }
+        return qs.filter(**filter_kwargs).distinct()
```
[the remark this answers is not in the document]
nemesifier: Valid. This duplicates security-sensitive parent scoping from `BaseEmailView`. Please extract the shared methods into a small mixin so the two implementations cannot drift.

openwisp_users/api/views.py
```diff
@@ -252,6 +253,72 @@ def get_object(self):
...
+
+    def get_serializer_context(self):
+        if getattr(self, "swagger_fake_view", False):
+            return None
+        context = super().get_serializer_context()
+        context["user"] = self.get_parent_queryset().first()
+        return context
```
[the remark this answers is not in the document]
nemesifier: I do not think this claim is correct. DRF serializers accept `context=None`, and schema generation does not evaluate this field queryset here. The real schema issue is the `KeyError` in `OrganizationMembershipListCreateView.get_queryset()`, which I reported separately.

openwisp_users/tests/test_api/test_views.py
```diff
@@ -32,6 +32,13 @@ def test_invalid_uuid_routes_return_404(self):
             "/api/v1/users/user/not-a-uuid/password/",
             "/api/v1/users/user/not-a-uuid/email/",
             "/api/v1/users/user/not-a-uuid/email/1/",
+            "/api/v1/users/user/not-a-uuid/organization-membership/",
+            "/api/v1/users/user/not-a-uuid/organization-membership/not-a-uuid/",
+            (
+                "/api/v1/users/user/"
+                "0c09d8e3-9e1d-4d2b-9a8c-6f1b2c3d4e5f/"
+                "organization-membership/not-a-uuid/"
+            ),
```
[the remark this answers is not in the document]
nemesifier: Valid. Build the valid membership detail URL with `reverse()` and replace only the organization UUID segment. The current hardcoded path will drift if the route prefix changes.

docs/user/rest-api.rst
```diff
@@ -203,6 +203,53 @@ Remove Email Address
...
+
+    PATCH /api/v1/users/user/{id}/organization-membership/{org_id}/
+
+Remove Organization Membership
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. code-block:: text
+
+    DELETE /api/v1/users/user/{id}/organization-membership/{org_id}/
+
```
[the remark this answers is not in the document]
nemesifier: Valid. Please document the actual permission contract: superusers can manage all memberships, while organization managers need the corresponding model permission, are limited to organizations they manage, and cannot manage superusers.

openwisp_users/tests/test_api/test_api.py
```diff
@@ -504,6 +515,233 @@ def test_delete_email_api(self):
...
+        self._create_org_user(organization=org1, user=org1_manager, is_admin=True)
+        administrator = Group.objects.get(name="Administrator")
+        org1_manager.groups.add(administrator)
+        self._create_org_user(organization=org1, user=self._get_admin(), is_admin=False)
+        self.client.force_login(org1_manager)
+        admin = self._get_admin()
+        path = reverse("users:organization_membership_list", args=(admin.pk,))
+        with self.assertNumQueries(4):
+            r = self.client.get(path)
+        self.assertEqual(r.status_code, 404)
```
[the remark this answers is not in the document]
nemesifier: Valid. Please add both managed and unmanaged manager DELETE cases. The same coverage gap exists for update, so include managed and cross-organization update cases as well.

openwisp_users/tests/test_api/test_api.py
```diff
@@ -504,6 +515,233 @@ def test_delete_email_api(self):
...
+    def test_organization_membership_patch_api(self):
+        user1 = self._create_user(username="user1", email="user1@email.com")
+        org1 = self._create_org(name="org1")
+        self._create_org_user(user=user1, organization=org1)
+        path = reverse("users:organization_membership_detail", args=(user1.pk, org1.pk))
+        data = {"is_admin": True}
+        with self.assertNumQueries(7):
+            r = self.client.patch(path, data, content_type="application/json")
+        self.assertEqual(r.status_code, 200)
+        self.assertTrue(r.data["is_admin"])
```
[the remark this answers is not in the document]
nemesifier: Valid. These tests differ only by the HTTP method, so they should be one test with `subTest` as required by the repository guidelines.

openwisp_users/api/views.py
```diff
@@ -252,6 +253,72 @@ def get_object(self):
...
+        filter_kwargs = {
+            "organization_id": self.kwargs["org_id"],
+        }
+        obj = get_object_or_404(queryset, **filter_kwargs)
+        self.check_object_permissions(self.request, obj)
+        return obj
+
+    def update(self, request, *args, **kwargs):
+        kwargs["partial"] = True
+        return super().update(request, *args, **kwargs)
```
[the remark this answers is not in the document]
nemesifier: Valid. `PUT` currently behaves exactly like a partial update, while both the documentation and generated schema imply a full representation. I would prefer standard PUT semantics. If partial PUT is intentional, please document it and make the generated schema match.

BHARATH0153:
@nemesifier done please check Thanks!

openwisp_users/api/serializers.py
```diff
@@ -213,6 +213,46 @@ def to_internal_value(self, data):
...
+            if (
+                organization is not None
+                and self.Meta.model.objects.filter(
+                    user=data["user"], organization=organization
+                ).exists()
+            ):
+                raise serializers.ValidationError(
+                    {
+                        "organization": _(
+                            "The user is already a member of this organization."
```
nemesifier, line 238: wouldn't the existing validation logic handle this?

openwisp_users/api/serializers.py
```diff
@@ -213,6 +213,46 @@ def to_internal_value(self, data):
...
+                        )
+                    }
+                )
+        elif data.get("organization") is not None and (
+            data["organization"].pk != self.instance.organization_id
+        ):
+            raise serializers.ValidationError(
+                {
+                    "organization": _(
+                        "Organization in the request body does not match the "
```
nemesifier: How can this happen in practice? Can you give me an example?

openwisp_users/api/serializers.py
BHARATH0153: The exclude_validation = ("user", "organization") skips the unique_together check in full_clean(). Without this explicit check, a duplicate create would result in a 500 IntegrityError instead of a clean 400 error.

openwisp_users/api/serializers.py
BHARATH0153: This can't really happen in normal usage. I'll make organization read-only on the detail endpoint and remove this check.

docs/user/rest-api.rst
```diff
@@ -272,6 +272,53 @@ Remove Email Address
 
     DELETE /api/v1/users/user/{id}/email/{id}/
 
+List Organization Memberships
```
nemesifier, line 275: The new API documentation lists the paths but omits their authorization rules. A client cannot tell that superusers have global access, while organization managers need the relevant model permission, are limited to organizations they manage, and cannot manage superusers.

Please add a concise note documenting these permission and scope restrictions.

Severity: P3

openwisp_users/api/serializers.py
```diff
@@ -213,6 +213,40 @@ def to_internal_value(self, data):
...
+                            "The user is already a member of this organization."
+                        )
+                    }
+                )
+        return super().validate(data)
+
+
+class OrganizationMembershipDetailSerializer(OrganizationMembershipSerializer):
+    class Meta(OrganizationMembershipSerializer.Meta):
+        extra_kwargs = {"organization": {"read_only": True}}
```
nemesifier, line 247: The detail serializer does not make `organization` read-only. DRF preserves the field explicitly declared on the parent serializer, so this `extra_kwargs` entry has no effect. I reproduced that a manager who manages both organizations can move an owner membership from `org1` to `org2`. The `OrganizationOwner` row still names `org1`, while its linked membership belongs to `org2`, and the user is then treated as `org2`'s owner.

Please explicitly declare the detail field as read-only, or keep it writable and reject a value that differs from the URL organization. Add a regression test for an owner membership as well.

Severity: P1

openwisp_users/api/views.py
```diff
@@ -331,6 +335,83 @@ def get_object(self):
...
```
nemesifier, line 406: This turns `PUT` into a partial update, so `PUT` accepts `{"is_admin": true}` without a complete representation. It differs from the other update endpoints and removes the route-body invariant needed for this membership resource.

Please restore standard DRF `PUT` semantics, keep `PATCH` for partial changes, and cover the distinction with a regression test.

Severity: P2

openwisp_users/api/views.py
```diff
@@ -331,6 +335,83 @@ def get_object(self):
         return obj
 
 
+class BaseOrganizationMembershipView(ProtectedAPIMixin, FilterByParent, GenericAPIView):
```
nemesifier, line 338: The parent lookup and organization-scoping logic duplicates `BaseEmailView`. This is authorization code, so maintaining two copies risks a tenant-isolation fix landing in one endpoint but not the other.

Please move the shared parent-scoping and serializer-context behavior into a common base or mixin, then reuse it from both endpoint families.

Severity: P3

openwisp_users/tests/test_api/test_api.py
```diff
@@ -530,6 +541,230 @@ def test_delete_email_api(self):
...
+        org1 = self._create_org(name="org1")
+        self._create_org_user(user=user1, organization=org1)
+        path = reverse("users:organization_membership_detail", args=(user1.pk, org1.pk))
+        with self.assertNumQueries(4):
+            r = self.client.get(path)
+        self.assertEqual(r.status_code, 200)
+        self.assertEqual(r.data["organization"], org1.pk)
+        self.assertFalse(r.data["is_admin"])
+
+    def test_organization_membership_put_api(self):
```
nemesifier, line 597: The `PUT` and `PATCH` tests duplicate the same setup, request payload, query budget, and assertions. They differ only by HTTP method, while `AGENTS.md` explicitly asks for these cases to share setup through `subTest`.

Please merge them into one update test that iterates over the two methods with `subTest`, resetting the membership state for each case.

Severity: P3

openwisp_users/tests/test_api/test_api.py
```diff
@@ -530,6 +541,230 @@ def test_delete_email_api(self):
...
+            username="org1_manager", email="org1_manager@test.com"
+        )
+        self._create_org_user(organization=org1, user=org1_manager, is_admin=True)
+        administrator = Group.objects.get(name="Administrator")
+        org1_manager.groups.add(administrator)
+        org2_user = self._create_user(username="org2_user", email="org2_user@test.com")
+        self._create_org_user(organization=org2, user=org2_user)
+        self.client.force_login(org1_manager)
+
+        with self.subTest("list memberships of a user outside managed organizations"):
```
nemesifier, line 733: The organization-manager tests exercise list, create, and retrieve scoping, but do not exercise either mutation on the detail endpoint. `PATCH` and `DELETE` are the operations that can alter or remove another organization's membership, so their tenant boundary remains unprotected by regression coverage.

Please add coverage for:
- a managed-organization `PATCH` that succeeds and persists the change
- an unmanaged-organization `PATCH` that returns `404` and preserves the record
- a managed-organization `DELETE` that returns `204`
- an unmanaged-organization `DELETE` that returns `404` and preserves the record

Severity: P2

openwisp_users/tests/test_api/test_api.py
```diff
@@ -530,6 +541,230 @@ def test_delete_email_api(self):
...
+        org1_user = self._create_user(username="org1_user", email="org1_user@test.com")
+        self._create_org_user(organization=org1, user=org1_user)
+        self.client.force_login(org1_manager)
+        path = reverse("users:organization_membership_list", args=(org1_user.pk,))
+        data = {"organization": org2.pk, "is_admin": True}
+        with self.assertNumQueries(12):
+            r = self.client.post(path, data, content_type="application/json")
+        self.assertEqual(r.status_code, 201)
+        self.assertEqual(r.data["organization"], org2.pk)
+        self.assertTrue(r.data["is_admin"])
```
nemesifier, line 699: This only checks the serialized response after the manager creates the membership. It does not independently verify that the membership was persisted for the requested organization.

Please assert that the saved `OrganizationUser` belongs to `org2` as well as checking the response.

Severity: P3

openwisp_users/tests/test_api/test_views.py
```diff
@@ -70,13 +70,26 @@ def test_invalid_uuid_routes_return_404(self):
...
```
nemesifier, line 79: The new invalid membership-route cases hardcode the `/api/v1/` prefix. This test will drift if the users API is mounted elsewhere, contrary to the repository rule to build internal URLs with named routes.

Please build a valid membership-detail URL with `reverse()` and replace only the UUID segment being tested with `not-a-uuid`.

Severity: P3

nemesifier (changes requested):
See my comments below

